PR #7, opened 2015-04-29
(MODULES-1957) use original MOF gem

ferventcoder:
Blocked by #5.

Switch over to using original MOF parser from ruby gems.

Use WMI style instead of CIM.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -25,7 +25,7 @@ gem 'librarian-puppet', '>=1.0.2'
 group :build do
   gem 'librarian-repo', :git => 'https://github.com/msutter/librarian-repo.git'
   gem 'cim'
-  gem 'mof', :git => 'https://github.com/msutter/mof.git'
+  gem 'mof'
   gem 'charlock_holmes'
 end
 
--- a/build/dsc/mof.rb
+++ b/build/dsc/mof.rb
@@ -42,7 +42,7 @@ module Dsc
       create_index_mof(@dsc_base_mof, base_mof_file_pathes)
 
       moffiles, options = MOF::Parser.argv_handler "moflint", []
-      options[:style] ||= :cim;
+      options[:style] ||= :wmi
       options[:includes] ||= []
       options[:quiet] ||= false
 


Iristyle:
I'm not sure I totally understand the `:wmi` vs `:cim` bit in f4fc491 - could you explain in a little more detail what the differences are?

https://github.com/kkaempf/mof/search?utf8=%E2%9C%93&q=style

ferventcoder:
> I'm not sure I totally understand the :wmi vs :cim bit in f4fc491 - could you explain in a little more detail what the differences are?

Don't feel that you are alone in this trying to understand something that is not super well-documented.

But note that if you are using iconv, apparently the helper switches you over to wmi anyways. https://github.com/kkaempf/mof/blob/4bad8309c2cb453dd99ecbc5e6af99d2e3bd1cdf/lib/mof/helper.rb#L88

Iristyle:
Yeah, interesting -- so it sniffs the file and turns on the `iconv` setting based on the byte order mark.

But I still think we should at least put *something* in the commit message about the `why` we needed to make the change / `what` it actually does.

I'm guessing it's related to what's mentioned in the projects [README](https://github.com/kkaempf/mof#featuresproblems) - https://msdn.microsoft.com/en-us/library/aa823192(VS.85).aspx

ferventcoder:
> But I still think we should at least put something in the commit message about the *why* we needed to make the change / *what* it actually does.

I agree with this. It works, but not really sure what it does differently other than how it evaluates MOF files (and that is still a bit fuzzy to me on how it does that differently).

ferventcoder:
Most of that is b/c there is very little documentation and the code appears to be abandonware. It's a risk factor to keep it around long term for sure. That's why I've been evaluating replacing it completely.

Iristyle:
Can we just add a little bit more info to the commit msg @ferventcoder so that we can get this guy merged?

Maybe try to generate with `:cim` and `:wmi` and compare the output or something?

ferventcoder:
> Maybe try to generate with :cim and :wmi and compare the output or something?

There is actually no difference in the output. I ran both of them and there were no visible changes in the underlying code. I think the real difference is in how it parses the MOF files.

Iristyle:
Ok - playing devil's advocate a bit here - if the output isn't any different, then why did we need to change it?  I'm a little confused...

ferventcoder:
We are making it more explicit.

PR #11, opened 2015-06-17
(FM-2562) Vendor wave kit

ferventcoder:
Add DSC Resource Wave Kit files into a vendor folder. We must start
with the ones that match what we can currently produce based on the
fact that we are using a MOF parser that doesn't work with MOF files as
produced by default from DSC Resources. Because of this, we are
vendoring the resource wave kit that the types are built on top of.
Once we have a valid replacement, we can update this folder with latest
components.

Add the vendored path to the PSModulePath by symlinking the vendored
folder into the program files modules path where all custom modules are
vendored. Without this change, we are not able to vendor the modules and
make it easy for folks to start using the module with little more than
WMF 5 installed.

Use a different output path for the resource generated MOF. Overwriting and
deleting the system32\Configuration MOFs messes up PowerShell DSC state in
WMF5 where it may have been okay in WMF4.

Import resources with ModuleName and ModuleVersion since they should match
exactly what we are vendoring, except for the possibility of items from the
PSDesiredStateConfiguration module. We have less control over the version
that comes with the built-in resources so only import by ModuleName.


ferventcoder:
This is getting closer

ferventcoder:
This is ready

vendor/DSCResources/README.md
```diff
@@ -0,0 +1,32 @@
+## DSC Resource Kit Wave 6
+
+This is a fixed version of the [DSC Resource Kit Wave 6](http://gallery.technet.microsoft.com/scriptcenter/DSC-Resource-Kit-All-c449312d)
+
+The fixes are for compatibility with the MOF parser. This is originally from https://github.com/msutter/dsc-resource-kit-wave-6
```
Iristyle: I don't understand the commentary... I could codegen everything in the lastest set of resources at https://github.com/PowerShell/DSCResources without making any modification to the MOFs.

Are we noting this temporarily, until we do a mass update of MOF + generated types to catch up to what's presently on GitHub?

ferventcoder: Yup, temporarily. The mass update is https://tickets.puppetlabs.com/browse/FM-2564

Iristyle:
@ferventcoder (aside from the issue that we're debugging), I think we might want to make sure all the vendored files are something Git better understands (i.e. not UCS-2), so that we can do proper diffs, etc.  We can add a `.gitattributes` to force diffing on all `psd1`, `psm1`, etc -- but there are some issues with that.

ferventcoder:
@Iristyle Short term anyway, we won't have control over these files soon as we switch.

Iristyle:
I'm not sure I understand what you mean about the `switch`. 

We do have the ability to control what goes in our repo, and we can propose upstream file encoding changes if we tire of re-encoding after grabbing what's available at https://github.com/PowerShell/DscResources

FWIW, I think it would be nice to be able to easily see diffs on GitHub / in local tools / etc.

ferventcoder:
Lost all the comments from https://github.com/ferventcoder/puppet-dsc/commit/a82d1b4673113cf3dcc31041ddb749e42205f178#commitcomment-11918390

ferventcoder:
Especially this guy - https://github.com/ferventcoder/puppet-dsc/commit/a82d1b4673113cf3dcc31041ddb749e42205f178#commitcomment-11918128

Iristyle:
FYI @ferventcoder - my branch is at https://github.com/Iristyle/puppetlabs-dsc/commits/pr/11

Still a handful of things left to do, but I've already squashed a ton of stuff down to that original commit to fix file encodings, etc.  Still evaluating the errantly encoded files that are in subsequent commit.

PR #12, opened 2015-06-17
(FM-2783) Fix Module Failure on WMF5 base resources

ferventcoder:
If a user tries to run the module on a Windows system running WMF 5
errors will occur when applying manifests with DSC resources specified:

~~~
Error: /Stage[main]/Main/Dsc_file[tmp_folder]: Could not
evaluate: WARNING: The configuration 'PuppetConfiguration' is loading
one or more
built-in resources without explicitly importing associated modules. Add
Import-DscResource -ModuleName 'PSDesiredStateConfiguration' to your
configuration to avoid this message.
False
~~~

Without this fix the warning will continue to occur producing nothing
to move forward with in setting the configuratin.

### DIFF
--- a/lib/puppet_x/msutter/templates/set_dsc_configuration.erb
+++ b/lib/puppet_x/msutter/templates/set_dsc_configuration.erb
@@ -6,10 +6,8 @@ Remove-Item $LCMConfigFolder\*.mof | out-null
 
 configuration PuppetConfiguration
 {
-<% unless ['msft_filedirectoryconfiguration'].include?(resource[:dscmeta_resource_name].downcase) -%>
 <%   if resource.parameters[:dscmeta_module_name] && resource[:dscmeta_import_resource] -%>
-    Import-DscResource <%= "-Name #{resource[:dscmeta_resource_name]}" if resource.parameters[:dscmeta_resource_name] %> <%= "-ModuleName #{resource[:dscmeta_module_name]}" if resource.parameters[:dscmeta_module_name] %>
-<%   end -%>
+    Import-DscResource -ModuleName <%= resource[:dscmeta_module_name] %>
 <% end -%>
 
     node "localhost"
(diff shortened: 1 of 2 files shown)


lib/puppet_x/msutter/templates/set_dsc_configuration.erb
```diff
@@ -10,6 +11,10 @@ configuration PuppetConfiguration
 <%   if resource.parameters[:dscmeta_module_name] && resource[:dscmeta_import_resource] -%>
     Import-DscResource <%= "-Name #{resource[:dscmeta_resource_name]}" if resource.parameters[:dscmeta_resource_name] %> <%= "-ModuleName #{resource[:dscmeta_module_name]}" if resource.parameters[:dscmeta_module_name] %>
 <%   end -%>
+<% else  -%>
+    if ($Posh5OrGreater ) {
+        Import-DscResource -ModuleName PSDesiredStateConfiguration
+    }
```
ferventcoder: This looks like it would only get applied for file, not all of the PSDesiredStateConfiguration files

ferventcoder: The rest are included in the regular call, so this is good as is.

Iristyle: What happens if we make this call in PS < 5? (i.e. do we need to guard it?)

ferventcoder: We have other fixes that we've just determined don't require the guarding at all. So yay for less stuff!

ferventcoder: The guard was there originally b/c the file resource

ferventcoder: The original guard was there (before this PR) b/c the file resource had issue importing both name and modulename, but there is no reason to import by name

PR #13, opened 2015-06-23
(FM-2916) Fix Gemfile Issue with master_manipulator

cowofevil:
The "master_manipulator" gem is private so the Gemfile should only load the gem
when the private mirror is available.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -38,13 +38,15 @@ group :test do
 end
 
 beaker_version = ENV['BEAKER_VERSION']
-group :system_tests do
-  if beaker_version
-    gem 'beaker', *location_for(beaker_version)
-  else
-    gem 'beaker', '~> 2.12.0'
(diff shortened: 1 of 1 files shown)


Gemfile
```diff
@@ -38,13 +38,15 @@ group :test do
...
+unless ENV['GEM_SOURCE'].nil?
+  group :system_tests do
+    if beaker_version
+      gem 'beaker', *location_for(beaker_version)
+    else
+      gem 'beaker', '~> 2.12'
+    end
+    gem 'master_manipulator', '~> 1.0'
   end
-  gem 'master_manipulator', '~> 1.0.0'
```
ferventcoder: I'm good with this - I'm guessing master_manipulator is required for beaker tests.

ferventcoder:
Thanks!

PR #17, opened 2015-07-02
(MODULES-1960) Invoke-DscResource

jpogran:
~~Depends on #11 to be pulled in first.~~

### DIFF
--- a/build/dsc/templates/dsc_type_spec.rb.erb
+++ b/build/dsc/templates/dsc_type_spec.rb.erb
@@ -162,26 +162,20 @@ describe Puppet::Type.type(:dsc_<%= resource.friendlyname.downcase %>) do
 
     describe "when dscmeta_import_resource is true (default) and dscmeta_module_name existing/is defined " do
 
-      it "should compute powershell dsc test script with Import-DscResource" do
-        expect(@provider.ps_script_content('test')).to match(/Import-DscResource/)
+      it "should compute powershell dsc test script with Invoke-DscResource" do
+        expect(@provider.ps_script_content('test')).to match(/Invoke-DscResource/)
       end
 
-      it "should compute powershell dsc set script with Import-DscResource" do
-        expect(@provider.ps_script_content('set')).to match(/Import-DscResource/)
+      it "should compute powershell dsc test script with method Test" do
+        expect(@provider.ps_script_content('test')).to match(/Method = \"test\"/)
       end
 
-    end
-
-    describe "when dscmeta_import_resource is false" do
-
-      it "should compute powershell dsc test script without Import-DscResource" do
-        dsc_<%= resource.friendlyname.downcase %>[:dscmeta_import_resource] = false
-        expect(@provider.ps_script_content('test')).not_to match(/Import-DscResource/)
+      it "should compute powershell dsc set script with Invoke-DscResource" do
+        expect(@provider.ps_script_content('set')).to match(/Invoke-DscResource/)
       end
 
-      it "should compute powershell dsc set script without Import-DscResource" do
-        dsc_<%= resource.friendlyname.downcase %>[:dscmeta_import_resource] = false
-        expect(@provider.ps_script_content('set')).not_to match(/Import-DscResource/)
+      it "should compute powershell dsc test script with method Set" do
+        expect(@provider.ps_script_content('set')).to match(/Method = \"set\"/)
       end
 
     end
--- a/lib/puppet/provider/base_dsc/powershell.rb
+++ b/lib/puppet/provider/base_dsc/powershell.rb
@@ -28,7 +28,7 @@ EOT
   def create
     Puppet.debug "\n" + ps_script_content('set')
     output = powershell(powershell_args, ps_script_content('set'))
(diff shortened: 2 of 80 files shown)


Iristyle:
I believe we're going to want some basic tests like those in place that use `ps_script_content`.

See https://github.com/puppetlabs/puppetlabs-dsc/blob/master/build/dsc/templates/dsc_type_spec.rb.erb#L163-L238

lib/puppet_x/msutter/helpers/dsc_provider_helpers.rb
```diff
@@ -61,6 +61,8 @@ def native_path(path)
           path.gsub(File::SEPARATOR, File::ALT_SEPARATOR)
         end
 
+        def dsc_invoke_method
```
Iristyle: Do we need this defined in the helper, or can this be local state set in the creation of the templated Powershell?

lib/puppet/provider/base_dsc/powershell.rb
```diff
@@ -60,5 +60,10 @@ def ps_script_content(mode)
     template.result(binding)
   end
 
-
+  def ps_invoke_content(mode)
```
Iristyle: @jpogran I believe you mentioned that we could just reuse `ps_script_content` given we're fully on board with using `Invoke-DscResource` now... but it looks like `ps_invoke_content` is still kicking around.

lib/puppet_x/msutter/templates/invoke_dsc_resource.erb
```diff
@@ -0,0 +1,74 @@
...
+$result = Invoke-DscResource @invokeParams
+
+switch($Method){
+  'Test'{
+    # true if what is expected
+    # false if it is not what is expected
+    $returnValue = [bool]($result.InDesiredState)
+  }
+  'Set'{
+    if($result.ReturnValue -eq $Null){
```
Iristyle: Simplify?

```powershell
$returnValue = @($Null, 0) -contains $result.ReturnValue
```

lib/puppet_x/msutter/templates/invoke_dsc_resource.erb
```diff
@@ -0,0 +1,74 @@
...
+<% end -%>
+}
+
+$result = Invoke-DscResource @invokeParams
+
+switch($Method){
+  'Test'{
+    # true if what is expected
+    # false if it is not what is expected
+    $returnValue = [bool]($result.InDesiredState)
```
Iristyle: This still seems a little odd - did we get to the bottom of what this value is supposed to be "officially"?

lib/puppet_x/msutter/templates/invoke_dsc_resource.erb
```diff
@@ -0,0 +1,74 @@
...
+    }elseif($result.ReturnValue -eq 0){
+      $returnValue = $true
+    }else{
+      $returnValue = $false
+    }
+  }
+}
+
+if($currentRefreshMode){
+  Set-RefreshMode $currentRefreshMode
```
Iristyle: If anything fails above, then we won't reset the LCM to it's initial state, which I think is problematic.

lib/puppet_x/msutter/templates/invoke_dsc_resource.erb
```diff
@@ -0,0 +1,74 @@
...
+          Settings
+          {
+              RefreshMode = $mode
+          }
+      }
+  }
+
+  LCMSettings -OutputPath $configOutputPath
+
+  Set-DscLocalConfigurationManager -Path $configOutputPath
```
Iristyle: What happens when this call fails (i.e. if there is another operation in progress)?

build/dsc/templates/dsc_type_spec.rb.erb
```diff
@@ -162,26 +162,20 @@ describe Puppet::Type.type(:dsc_<%= resource.friendlyname.downcase %>) do
 
     describe "when dscmeta_import_resource is true (default) and dscmeta_module_name existing/is defined " do
 
-      it "should compute powershell dsc test script with Import-DscResource" do
-        expect(@provider.ps_script_content('test')).to match(/Import-DscResource/)
+      it "should compute powershell dsc test script with Invoke-DscResource" do
+        expect(@provider.ps_invoke_content('test')).to match(/Invoke-DscResource/)
```
Iristyle: Ditto changes re: `ps_script_content` vs `ps_invoke_content`

lib/puppet/provider/base_dsc/powershell.rb
jpogran: It wasn't clear to me whether we reached consensus to keep `ps_script_content` or not. If that's the case, then I'll change back to that

lib/puppet_x/msutter/helpers/dsc_provider_helpers.rb
jpogran: Nope. Oversight. Removing now.

ferventcoder:
`\o/`

PR #22, opened 2015-07-09
(FM-2564) Update Generated Types

Iristyle:


ferventcoder:
Merged in #43, I take it this is next on the list?

Iristyle:
@ferventcoder / @jpogran - a couple of notes here on what I need to do to close this out.

* I want to add a rake task to clone down the types / refresh the vendor dir - should be pretty easy
* I think anything that looks wrong in b8df8a9 we want to look at fixing the upstream repos.
* As part of 0d3e5eb I found that `MSFT_xDSCWebService.Schema.mof` still had a BOM, which killed MOF generation - we'll want to fix the upstream repo.

We could also get up through 6e54f96 merged, and do the remaining updates in a subsequent PR to at least move some of this forward for the sake of @zreichert / @cowofevil testing.

Iristyle:
FYI rolled back in time to 6-22 modules so that @cowofevil can merge up through 6e54f96 so that we can do the ticket around moving vendored modules.

Then I can get us up a later drop of what's on GitHub

PR #29, opened 2015-07-20
(MODULES-1960) Don't version in-OS resources

Iristyle:
- Fix a bug in the logic around when to specify a module version when
   generating PowerShell code.
 - Also, after some discussion, decided to use the presence / absence
   of dscmeta_module_version to determine whether to use module name or
   module name + version. In the code generation for specific types - in
   this case PSDesiredStateConfiguration - omit dscmeta_module_version.

   After some discussion it was decided since these resources are not
   vendored, and can vary based on the operating system version that the
   Puppet module is installed on, that we should be less strict around
   requesting a particular version.  It's possible that minor .z
   versions will have the same interface, but may include OS specific
   bugfixes, etc.  Hardcoding a single version number is too inflexible
   given these DSC types are not vendored.

### DIFF
--- a/lib/puppet_x/msutter/templates/invoke_dsc_resource.erb
+++ b/lib/puppet_x/msutter/templates/invoke_dsc_resource.erb
@@ -42,12 +42,13 @@ $invokeParams = @{
 <% dsc_parameters.each do |p| -%>
     <%= p.name.to_s.gsub(/^dsc_/,'') %> = <%= format_dsc_value(p.value) %>
 <% end -%>
-  }
-<% if resource[:dscmeta_module_name].casecmp('PSDesiredStateConfiguration') == 0 -%>
+  }<% if resource.parameters[:dscmeta_module_version] %>
   ModuleName = @{
-    <%= "ModuleName=\"#{resource[:dscmeta_module_name]}\"" %>
-    <%= "RequiredVersion=\"#{resource[:dscmeta_module_version]}\"" if resource.parameters[:dscmeta_module_version] %>
+    ModuleName      = <%= "\"#{resource[:dscmeta_module_name]}\"" %>
+    RequiredVersion = <%= "\"#{resource[:dscmeta_module_version]}\"" %>
   }
+<% else %>
+  ModuleName = <%= "\"#{resource[:dscmeta_module_name]}\"" %>
 <% end -%>
 }
 
--- a/build/dsc/templates/dsc_type.rb.erb
+++ b/build/dsc/templates/dsc_type.rb.erb
@@ -44,11 +44,11 @@ Puppet::Type.newtype(:dsc_<%= resource.friendlyname.downcase %>) do
   newparam(:dscmeta_module_name) do
     defaultto "<%= resource.ps_module.name %>"
   end
-
+<%  if resource.ps_module.name != 'PSDesiredStateConfiguration' %>
   newparam(:dscmeta_module_version) do
     defaultto "<%= resource.ps_module.version %>"
   end
-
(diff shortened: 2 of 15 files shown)


lib/puppet_x/msutter/templates/invoke_dsc_resource.erb
```diff
@@ -42,12 +42,13 @@ $invokeParams = @{
 <% dsc_parameters.each do |p| -%>
     <%= p.name.to_s.gsub(/^dsc_/,'') %> = <%= format_dsc_value(p.value) %>
 <% end -%>
-  }
-<% if resource[:dscmeta_module_name].casecmp('PSDesiredStateConfiguration') == 0 -%>
```
ferventcoder: I'm guessing this for some reason wasn't always catching the name for PSDesiredStateConfiguration?

ferventcoder: I'm thumbs up on moving the check back to generating the type and simplifying things a bit.

lib/puppet_x/msutter/templates/invoke_dsc_resource.erb
```diff
@@ -42,12 +42,13 @@ $invokeParams = @{
...
+  }<% if resource.parameters[:dscmeta_module_version] %>
   ModuleName = @{
-    <%= "ModuleName=\"#{resource[:dscmeta_module_name]}\"" %>
-    <%= "RequiredVersion=\"#{resource[:dscmeta_module_version]}\"" if resource.parameters[:dscmeta_module_version] %>
+    ModuleName      = <%= "\"#{resource[:dscmeta_module_name]}\"" %>
+    RequiredVersion = <%= "\"#{resource[:dscmeta_module_version]}\"" %>
   }
+<% else %>
+  ModuleName = <%= "\"#{resource[:dscmeta_module_name]}\"" %>
 <% end -%>
```
ferventcoder: Out of curiosity, why do we need and if else then if we can cover that line in the inline if statement? Is it the extra empty line in ModuleName?

e.g.

```powershell
ModuleName @{
   ModuleName = "somename"

}
```

Does it error when it's an array with only one element? That's what I'm guessing...

lib/puppet_x/msutter/templates/invoke_dsc_resource.erb
Iristyle: The problem was that it *should* have been `!= 0`

It just so happened to work in my testing because I didn't validate against `x` style resources.

So I chose to rewrite this a little bit to not hardcode the `PSDesiredStateConfiguration` here, but to use a different heuristic.

lib/puppet_x/msutter/templates/invoke_dsc_resource.erb
Iristyle: Yup, it errors with just a `ModuleName` in there... if you want to specify the parent `ModuleName` as a hash, both `ModuleName` and `RequiredVersion` are required.

ferventcoder: :+1:

lib/puppet_x/msutter/templates/invoke_dsc_resource.erb
ferventcoder: Makes sense, better heuristic as well.

ferventcoder: > The problem was that it should have been != 0

Missed that logic bug, but I see it now! Ha!

PR #28, opened 2015-07-17
(MODULES-2251) Update Tests for "Invoke-DscResource"

cowofevil:
Update the acceptance and integration tests. Also update the test run scripts
to use the correct pre-suites.

PR #31, opened 2015-07-21
(MODULES-2228) Create Tests for "Registry" DSC Resource

cowofevil:
Add automated tests for the "Registry" DSC resource type. Also, update the
"dsc_utils.rb" helper library to be more better-er.

Cannot merge this until #28 is merged.


zreichert:
This PR causes tests/acceptance/tests/basic_functionality/puppet_apply_debug_dsc_manifest.rb to fail.

PR #32, opened 2015-07-22
(DO NOT MERGE)(MODULES-2243) Remove LCM RefreshMode Change

jpogran:
Removes setting the LCM RefreshMode in the Invoke DSC Resource script and
have it bail out if it is set to anything else than Disabled.

Creates a Puppet type that will set the LCM RefreshMode to one of the
supported values

### DIFF
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -1,0 +1,6 @@
+fixtures:
+  forge_modules:
+    stdlib: "puppetlabs/stdlib"
+    powershell: "puppetlabs/powershell"
+  symlinks:
+    "dsc": "#{source_dir}"
--- a/lib/puppet_x/msutter/templates/invoke_dsc_resource.erb
+++ b/lib/puppet_x/msutter/templates/invoke_dsc_resource.erb
@@ -4,33 +4,10 @@ if (!(Test-Path($PuppetModulesFolder))) {
   & cmd.exe /c mklink /d "$PuppetModulesFolder" "<%= vendored_modules_path %>" | Out-Null
 }
 
-function Set-RefreshMode
-{
-  param($mode)
-
-  $configOutputPath = [IO.Path]::GetTempPath()
-
-  [DscLocalConfigurationManager()]
-  Configuration LCMSettings {
-      Node localhost
-      {
-          Settings
-          {
-              RefreshMode = $mode
-          }
-      }
-  }
-
-  LCMSettings -OutputPath $configOutputPath | Out-Null
-
-  Set-DscLocalConfigurationManager -Path $configOutputPath
-}
-
 $currentState = Get-DscLocalConfigurationManager
 
 if ($currentState.RefreshMode -ne 'Disabled') {
-  $currentRefreshMode = $currentState.RefreshMode
-  Set-RefreshMode 'Disabled'
+  throw "DSC LCM RefreshMode must be set to Disabled for Puppet to execute DSC Resources!"
 }
 
 $invokeParams = @{
--- a/manifests/lcm_config.pp
+++ b/manifests/lcm_config.pp
@@ -1,0 +1,12 @@
+define dsc::lcm_config (
+  $refresh_mode = 'Disabled'
+) {
+
+  validate_re($refresh_mode, '^(Disabled|Pull|Push)$', 'refresh_mode must be one of \'Disabled\', \'Push\', \'Pull\'')
+
+  exec { "dsc_provider_set_lcm_refreshmode_${refresh_mode}":
+    provider => 'powershell',
+    onlyif   => "if ((Get-DscLocalConfigurationManager).RefreshMode -eq '${refresh_mode}') {exit 1} else {exit 0}",
+    command  => template('dsc/set-lcm-mode.ps1.erb'),
+  }
+}
--- a/metadata.json
+++ b/metadata.json
@@ -37,6 +37,13 @@
     }
   ],
   "dependencies": [
-
+    {
+      "name": "puppetlabs-stdlib",
+      "version_requirement": ">= 1.0.0"
+    },
+    {
+      "name": "puppetlabs-powershell",
+      "version_requirement": ">= 1.0.1"
+    }
   ]
 }
--- a/templates/set-lcm-mode.ps1.erb
+++ b/templates/set-lcm-mode.ps1.erb
@@ -1,0 +1,50 @@
+$lcmRefreshMode = '<%= @refresh_mode %>'
+$generationDate = (get-date -format "M/d/yyyy HH:mm:ss")
+$encoding       = [Text.Encoding]::UTF8
+$namespace      = 'root/Microsoft/Windows/DesiredStateConfiguration'
+$className      = 'MSFT_DSCLocalConfigurationManager'
+$generationHost = [Environment]::MachineName
+
+$mofData = @"
+/*
+@TargetNode=''localhost''
+@GeneratedBy=PUPPET
+@GenerationDate=$generationDate
+@GenerationHost=$generationHost
+*/
+
+instance of MSFT_DSCMetaConfiguration as `$MSFT_DSCMetaConfiguration1ref
+{
+RefreshMode = "$lcmRefreshMode";
+};
+
+instance of OMI_ConfigurationDocument
+{
+Version="2.0.0";
+MinimumCompatibleVersion = "1.0.0";
+CompatibleVersionAdditionalProperties= { "MSFT_DSCMetaConfiguration:StatusRetentionTimeInDays" };
+Author="PUPPET";
+GenerationDate="$generationDate";
+GenerationHost="$generationHost";
+Name="LCMSettings";
+};
+"@
+
+$totalSize = [BitConverter]::GetBytes($mofData.Length + 4)
+$bytes     = $totalSize + $encoding.GetBytes($mofData)
+
+$cimClass       = Get-CimClass -Namespace $namespace -ClassName $className
+$sessionOptions = New-CimSessionOption -Protocol Wsman
+$wsmanSession   = New-CimSession -ComputerName localhost -SessionOption $sessionOptions
+
+$invokeParams = @{
+  Verbose    = $true
+  CimClass   = $cimClass
+  CimSession = $wsmanSession
+  MethodName = "SendMetaConfigurationApply"
+  Arguments  = @{
+    ConfigurationData = $bytes;
+  }
+}
+
+Invoke-CimMethod @invokeParams
--- a/spec/defines/lcm_config_spec.rb
+++ b/spec/defines/lcm_config_spec.rb
@@ -1,0 +1,47 @@
+require 'spec_helper'
+
+RSpec.describe 'dsc::lcm_config', :type => :define do
+
+  context 'refresh_mode should default to Disabled' do
+    let (:title) { 'lcm_default' }
+
+    it {
+      should contain_dsc__lcm_config("#{title}").with(
+        {
+          'refresh_mode' => 'Disabled'
+        }
+      )
+    }
+  end
+
+  ["Disabled","Push","Pull"].each do |mode|
+    context "when refresh_mode => #{mode}" do
+      let (:title) { "lcm_#{mode}" }
+      let(:params) { {:refresh_mode => "#{mode}"} }
+
+      it {should compile}
+
+      it {
+        should contain_dsc__lcm_config("#{title}").with(
+          {
+            'refresh_mode' => "#{mode}"
+          }
+        )
+      }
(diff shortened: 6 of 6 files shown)


lib/puppet_x/msutter/templates/invoke_dsc_resource.erb
```diff
@@ -4,33 +4,10 @@ if (!(Test-Path($PuppetModulesFolder))) {
...
-
-  Set-DscLocalConfigurationManager -Path $configOutputPath
-}
-
 $currentState = Get-DscLocalConfigurationManager
 
 if ($currentState.RefreshMode -ne 'Disabled') {
-  $currentRefreshMode = $currentState.RefreshMode
-  Set-RefreshMode 'Disabled'
+  throw "DSC LCM RefreshMode must be set to Disabled for Puppet to execute DSC Resources!"
```
Iristyle: Perhaps not something to address now, but I wonder if we can better propagate error messages if we return something structured to Ruby upon error.

Food for thought.

templates/set-lcm-mode.ps1.erb
```diff
@@ -0,0 +1,51 @@
+Write-Host "Setting LCM"
+$LCMRefreshMode = '<%= @refresh_mode %>'
+$generationDate = (get-date -format "M/d/yyyy HH:mm:ss")
+$encoding       = [system.Text.Encoding]::UTF8
+$Namespace      = 'root/Microsoft/Windows/DesiredStateConfiguration'
+$ClassName      = 'MSFT_DSCLocalConfigurationManager'
+
+Write-Verbose "LCM Mode: $LCMRefreshMode"
```
Iristyle: So FYI I don't think we have a way to pass in a `-Verbose` to PowerShell from Ruby...

templates/set-lcm-mode.ps1.erb
```diff
@@ -0,0 +1,51 @@
+Write-Host "Setting LCM"
+$LCMRefreshMode = '<%= @refresh_mode %>'
+$generationDate = (get-date -format "M/d/yyyy HH:mm:ss")
+$encoding       = [system.Text.Encoding]::UTF8
```
Iristyle: style nit - I usually omit the `system.` since it's not necessary

templates/set-lcm-mode.ps1.erb
```diff
@@ -0,0 +1,51 @@
+Write-Host "Setting LCM"
```
Iristyle: style nit - Single quotes `'` vs double quotes `"`

templates/set-lcm-mode.ps1.erb
jpogran: The param set at line 42 sets verbose for the Invoke-CimMethod at line 51, so verbose messages show up. Line 1 should be a verbose message as well instead of write-host, but both were added during debugging and really don't add much value otherwise so it may be best just to remove both.

templates/set-lcm-mode.ps1.erb
jpogran: Really? I have always followed shortened on the command line, fully qualified (full namespace, no aliases, etc...) in the script.

lib/puppet_x/msutter/templates/invoke_dsc_resource.erb
jpogran: I don't think we're going to be able to achieve that without doing the json passing we have been discussing.

templates/set-lcm-mode.ps1.erb
```diff
@@ -0,0 +1,51 @@
...
+$ClassName      = 'MSFT_DSCLocalConfigurationManager'
+
+Write-Verbose "LCM Mode: $LCMRefreshMode"
+
+$mofData = @"
+/*
+@TargetNode=''localhost''
+@GeneratedBy=PUPPET
+@GenerationDate=$generationDate
+@GenerationHost=PUPPET
```
jpogran: Was wondering if we should be setting this to the target node's hostname instead of Puppet since it's technically generated on the target node and not on puppet.

Iristyle: Yeah, I think so.  Not sure if that value get's written into any extended logging anywhere, but seems like a good idea.

lib/puppet_x/msutter/templates/invoke_dsc_resource.erb
Iristyle: Yeah, we should probably file a ticket for that, and take care of it next sprint.

metadata.json
```diff
@@ -33,10 +33,17 @@
     },
     {
       "name": "puppet",
-      "version_requirement": "3.x"
+      "version_requirement": ">= 3.4.0"
```
Iristyle: I would keep this specific line in another commit for bookkeeping purposes.

templates/set-lcm-mode.ps1.erb
Iristyle: Yeah, I tend to write code that same way - long form in scripts, except for a few exceptions.

* Omit `System.` since it's redundant and typically implied (which can help keep line length down)
* `%` instead of `ForEach-Object`
* `?` instead of `Where-Object`

I have a little style guide I wrote years ago around here somewhere.....

templates/set-lcm-mode.ps1.erb
```diff
@@ -0,0 +1,49 @@
...
+  Verbose    = $true
+  CimClass   = $cimClass
+  CimSession = $wsmanSession
+  MethodName = "SendMetaConfigurationApply"
+  Arguments  = @{
+    ConfigurationData = $bytes;
+  }
+}
+
+Invoke-CimMethod @invokeParams
```
cowofevil: Missing newline at end of file.

manifests/lcm_config.pp
```diff
@@ -0,0 +1,12 @@
...
+) {
+
+  validate_re($refresh_mode, '^(Disabled|Pull|Push)$', 'refresh_mode must be one of \'Disabled\', \'Push\', \'Pull\'')
+
+  exec { "dsc_provider_set_lcm_refreshmode_${refresh_mode}":
+    provider => 'powershell',
+    onlyif   => "if ((Get-DscLocalConfigurationManager).RefreshMode -eq '${refresh_mode}') {exit 1} else {exit 0}",
+    command  => template('dsc/set-lcm-mode.ps1.erb'),
+  }
+}
```
cowofevil: Missing newline at end of file.

metadata.json
ferventcoder: That was my fault.

ferventcoder: @jpogran and it will be 3.7.0 by the way :)

metadata.json
```diff
@@ -29,14 +29,17 @@
   "requirements": [
     {
       "name": "pe",
-      "version_requirement": "3.x"
-    },
-    {
-      "name": "puppet",
-      "version_requirement": "3.x"
+      "version_requirement": ">= 3.4.0"
```
ferventcoder: Pull this out completely. This will be best done with a different ticket.

ferventcoder: Plus it looks like it might have gotten a bit heavy handed and removed the puppet requirement.

Iristyle:
In 9dd238f we probably want a `.fixtures.yml` as well to get those deps installed with `rake spec_prep`

```yml
fixtures:
  forge_modules:
    powershell: "puppetlabs/powershell"
    stdlib: "puppetlabs/stdlib"
  symlinks:
    dsc: "#{source_dir}"
```

manifests/lcm_config.pp
```diff
@@ -0,0 +1,12 @@
+define dsc::lcm_config (
+  $refresh_mode = 'Disabled'
+) {
+
+  validate_re($refresh_mode, '^(Disabled|Pull|Push)$', 'refresh_mode must be one of \'Disabled\', \'Push\', \'Pull\'')
```
cowofevil: Not to quibble, but the regex is case sensitive and Windows people are use to insensitive casing. May want to expect upper and lower case.

ferventcoder: We haven't decided on that one yet. It depends on whether you can set the mode case insensitively or not.

.fixtures.yml
```diff
@@ -0,0 +1,6 @@
+fixtures:
+  forge_modules:
+    stdlib: "puppetlabs/stdlib"
+    powershell: "puppetlabs/powershell"
+  symlinks:
+    "dsc": "#{source_dir}"
```
Iristyle: Missing newline ;0

lib/puppet_x/msutter/templates/invoke_dsc_resource.erb
Iristyle: I have confirmed that we probably want to improve this error reporting, because the way it bails right now is pretty user-unfriendly.  Certainly good enough for this PR, but we have room for improvement in a future iteration.


```
Error: /Stage[main]/Main/Dsc_registry[registry_test]: Could not evaluate: Execution of 'C:\Windows\sysnative\WindowsPowershell\v1.0\powershell.exe -NoProfile -NonInteractive -NoLogo -ExecutionPolicy Bypass -Command $script:ErrorActionPreference = 'Stop'
$PuppetModulesFolder = Join-Path $env:ProgramFiles 'WindowsPowerShell\Modules\PuppetVendoredModules'
if (!(Test-Path($PuppetModulesFolder))) {
  & cmd.exe /c mklink /d "$PuppetModulesFolder" "c:\source\dsc\puppetlabs-dsc\spec\fixtures\modules\dsc\vendor\DSCResources" | Out-Null
}

$currentState = Get-DscLocalConfigurationManager

if ($currentState.RefreshMode -ne 'Disabled') {
  throw "DSC LCM RefreshMode must be set to Disabled for Puppet to execute DSC Resources!"
}

$invokeParams = @{
  Name          = 'Registry'
  Method        = 'test'
  ErrorAction   = 'Continue'
  ErrorVariable = 'invokeError'
  Property      = @{
    key = 'HKEY_LOCAL_MACHINE\SOFTWARE\TestKey'
    valuename = 'TestStringValue'
    valuedata = @('Cats')
    valuetype = 'String'
    ensure = 'Present'
  }
  ModuleName = "PSDesiredStateConfiguration"
}

$result = Invoke-DscResource @invokeParams

if ($currentRefreshMode) { Set-RefreshMode $currentRefreshMode }
if ($invokeError) { throw $invokeError }

switch ($invokeParams.Method) {
  'Test' {
    [bool]($result.InDesiredState)
  }
  'Set' {
    @($Null, 0) -contains $result.ReturnValue
  }
}
' returned 1: DSC LCM RefreshMode must be set to Disabled for Puppet to execute DSC Resources!
At line:10 char:3
+   throw "DSC LCM RefreshMode must be set to Disabled for Puppet to ex ...
+   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : OperationStopped: (DSC LCM Refresh... DSC Resources!:String) [], R
   untimeException
    + FullyQualifiedErrorId : DSC LCM RefreshMode must be set to Disabled for Puppet to execute
   DSC Resources!
```

Iristyle:
Overall this is really close to merge as of e17e680.  

I'm pretty satisfied with local tests with multiple resource manifests, like this one:

```puppet
dsc_service { 'w32time':
  dsc_name  => 'w32time',
  dsc_state => 'Running'
}

dsc_registry { 'registry_test_binary':
  dsc_ensure    => 'Present',
  dsc_key       => 'HKEY_LOCAL_MACHINE\SOFTWARE\TestKey',
  dsc_valuename => 'TestBinaryValue',
  dsc_valuedata => 'BEEF',
  dsc_valuetype => 'Binary',
}

dsc_registry { 'registry_test_dword':
  dsc_ensure    => 'Present',
  dsc_key       => 'HKEY_LOCAL_MACHINE\SOFTWARE\TestKey',
  dsc_valuename => 'TestDwordValue',
  dsc_valuedata => '1',
  dsc_valuetype => 'Dword',
}

dsc_registry { 'registry_test_string':
  dsc_ensure    => 'Present',
  dsc_key       => 'HKEY_LOCAL_MACHINE\SOFTWARE\TestKey',
  dsc_valuename => 'TestStringValue',
  dsc_valuedata => 'Cats',
  dsc_valuetype => 'String',
}

dsc_service { 'w32time2':
  dsc_name  => 'w32time',
  dsc_state => 'Stopped'
}
```

Let's get #38 in first since we're going to want the rake task `spec_prep` working properly on Windows.  Then I believe you have a WIP test to update this PR with @jpogran, and some minor cleanup based on comments, and I think we can :shipit:

PR #33, opened 2015-07-22
(MODULES-2263) Add Test for DSC Resources with Array

cowofevil:
Add a test that verifies that a user can specify an array for a DSC resource
parameter.


cowofevil:
@zreichert Please wait until #35 is merged before reviewing this PR.

PR #36, opened 2015-07-23
(maint) Enable rake without Bundler 'build' group

Iristyle:
- The 'build' gems cannot presently be installed on Windows due to
   some conflicts with 'charlock_holmes' that includes native code that
   doesn't compile on Windows, and 'librarian-repo' which relies on
   'rubyzip' that includes native code that doesn't compile on Windows.

   This makes a typical Windows bundle install:

   `bundle install --path .bundle/gems --without system_tests build`

   However, the current rake tasks always try to load these gems, and
   therefore rake tasks cannot run on Windows.

   At a future date, some of the build process will be reworked to
   run on Windows, but for now, simply guard the loading of build
   related gems based on their existence.

   Build tasks will not work properly on Windows, but at least other
   tasks like 'spec_prep', included with puppetlabs_spec_helper, are
   now functional.

### DIFF
--- a/build/dsc.rb
+++ b/build/dsc.rb
@@ -4,11 +4,15 @@ require 'fileutils'
 require 'pathname'
 require 'erb'
 require 'open-uri'
-require 'mof'
-require 'cim'
 require 'find'
 require 'logger'
-require 'charlock_holmes/string'
+
+# Bundler group 'build'
+if Bundler.rubygems.find_name('mof').any?
+  require 'mof'
+  require 'cim'
+  require 'charlock_holmes/string'
+end
 
 # Debug
 require 'pry' if Bundler.rubygems.find_name('pry').any?


build/dsc.rb
```diff
@@ -4,11 +4,15 @@
...
 require 'erb'
 require 'open-uri'
-require 'mof'
-require 'cim'
 require 'find'
 require 'logger'
-require 'charlock_holmes/string'
+
+# Bundler group 'build'
+if Bundler.rubygems.find_name('mof').any?
```
ferventcoder: Is the issue that it doesn't work on Windows? Why not shield by `if !Puppet::Util::Platform.windows?` instead?

ferventcoder: On second thought, that would need puppet required. Is there some other way to determine we are on Windows and shield this?

Iristyle: This is actually more of a 'what gems are installed' problem, than a 'Windows problem'.

It just so happens that Windows can't install the `build` gem group... but if you did a `bundle install --path .bundle/gems --without build` on any other OS, you'd have the same problem running rake tasks, which is why I opted to guard it this way.

ferventcoder: Fair enough - I'm good to move forward on this as a stop gap and come back to it later if we need to enhance it. :)

PR #37, opened 2015-07-23
(maint) puppetlabs_spec_helper ~>0.10.3 / rake to default group

ferventcoder:
Using PSH with versions lower than 0.9.0 results in issues
attempting to symlink on Windows. We'll take the latest
version as the dependency since it has the best fixes
available for cross platform testing.

Rake is required by other groups, so move it from just the
development group to the default group to allow it to always
be pulled in.

### DIFF
--- a/.sync.yml
+++ b/.sync.yml
@@ -13,6 +13,7 @@ Gemfile:
         version: '~>2.14.1'
       - gem: puppet-lint
       - gem: puppetlabs_spec_helper
+        version: '~>0.10.3'
       - gem: puppet_facts
       - gem: mocha
         version: '~>0.10.5'
--- a/.sync.yml
+++ b/.sync.yml
@@ -8,7 +8,6 @@ Gemfile:
   supports_windows: true
   required:
     ':development':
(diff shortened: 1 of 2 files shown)


ferventcoder:
It should automatically prefer 0.10.3 and looking at some appveyor logs of older builds it does, but this makes sure that we are using a known good version.

Iristyle:
Given my PSH issue was not in fact a PSH issue, but put into the removal of `librarian-puppet` as part of PR #38, I'm inclined to close this and revisit at another time.  Thoughts?

ferventcoder:
The modulesync will push this again later. It's not a big deal either way, but just ensuring that we don't run into issue with PSH later if for some reason bundler makes the decision to get an older version.

Iristyle:
You want to update this with latest modulesync @ferventcoder ?

ferventcoder:
@Iristyle this is updated. I also moved rake out to the default group as @cowofevil had started earlier.

jpogran:
Works on a fresh clone without error. Good to merge.

PR #38, opened 2015-07-23
(maint) Remove librarian-puppet

Iristyle:
- Librarian-puppet is completely unused in this module, so remove any
   references, including related rake tasks.  The existing rake tasks
   were in conflict with the usage of .fixtures.yml in
   puppetlabs_spec_helper (PSH), which is the default tool uesd by
   Puppet for modules.

   The conflicting task also included a call to `ln -s`, rendering the
   current approach incompatible with Windows. This commit restores the
   ability to correctly install fixtures with PSH on Windows.

### DIFF
--- a/Puppetfile
+++ b/Puppetfile
@@ -1,1 +1,0 @@
-forge "https://forgeapi.puppetlabs.com"
--- a/Rakefile
+++ b/Rakefile
@@ -19,15 +19,4 @@ PuppetLint.configuration.log_format = '%{path}:%{linenumber}:%{KIND}: %{message}
 PuppetLint.configuration.send("disable_class_inherits_from_params_class")
 PuppetLint.configuration.send("disable_80chars")
 
-# use librarian-puppet to manage fixtures instead of .fixtures.yml
-# offers more possibilities like explicit version management, forge downloads,...
-task :librarian_spec_prep do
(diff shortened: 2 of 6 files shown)


Puppetfile
```diff
@@ -1 +0,0 @@
-forge "https://forgeapi.puppetlabs.com"
```
ferventcoder: Boy that file had quite a bit in it

Rakefile
```diff
@@ -19,15 +19,4 @@ PuppetLint.configuration.log_format = '%{path}:%{linenumber}:%{KIND}: %{message}
 PuppetLint.configuration.send("disable_class_inherits_from_params_class")
 PuppetLint.configuration.send("disable_80chars")
 
-# use librarian-puppet to manage fixtures instead of .fixtures.yml
-# offers more possibilities like explicit version management, forge downloads,...
-task :librarian_spec_prep do
-  sh "librarian-puppet install --path=spec/fixtures/modules/"
-  pwd = `pwd`.strip
-  unless File.directory?("#{pwd}/spec/fixtures/modules/dsc")
-    sh "ln -s #{pwd} #{pwd}/spec/fixtures/modules/dsc"
```
ferventcoder: yep... :/

PR #39, opened 2015-07-24
(DO NOT MERGE)(MODULES-2267) Fix ensure absent

jpogran:
The method set_test_dsc_parameters resets any dsc_ensure line to
'present' if the user set dsc_ensure to 'absent'. When the destroy
method is called, set_test_dsc_parameters has already been called and
changed the dsc_ensure value to the opposite of what we want.

The exists? method as is, is a silent bug because it executes
set_test_dsc_parameters after the DEBUG output is shown to the user,
hiding the fact that it is changing the dsc_ensure value during the test
run, making it appear that it is changed during teh destroy run. Moving
it up one line shows the value change in the DEBUG output. Since the
values are changed with set_test_dsc_parameters and not set back, the
destroy method passes the 'present' value to the DSC Script instead of
the 'absent' value the user specified.

By invoking set_original_dsc_parameters before we compile the script in
the destroy method we set the values back to their original settings. We
still need to invoke set_original_dsc_parameters in the exists? method
because we need to flip our logic for DSC. DSC returns false if the
target node is not in the desired state, so we have to inverse our logic
to get it to give us a true (true meaning in this case that the thing we
want to destroy actually exists).

### DIFF
--- a/build/dsc/templates/dsc_type_spec.rb.erb
+++ b/build/dsc/templates/dsc_type_spec.rb.erb
@@ -242,6 +242,22 @@ describe Puppet::Type.type(:dsc_<%= resource.friendlyname.downcase %>) do
       end
 
     end
+
+    describe "when dsc_ensure is 'absent' in a real execution run that both tests and sets" do
+
+      it "should compute powershell dsc set script in which ensure value is 'absent'" do
+        dsc_<%= resource.friendlyname.downcase %>.original_parameters[:dsc_ensure] = 'absent'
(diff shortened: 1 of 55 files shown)


ferventcoder:
LGTM :+1:

Iristyle:
FYI we have a few changes to make to the approach prior to merge.

Iristyle:
The new approach is captured in my branch at https://github.com/Iristyle/puppetlabs-dsc/commit/c32affaac86a7615d9e27c6466a7cceba1327476

PR #42, opened 2015-07-29
(maint) remove unused MOF provider / templates

Iristyle:
This is a tech debt / unused code removal PR and should be pretty straightforward to merge.

 - Prior implementations of this module had spiked a MOF provider, but
   with the advent of Invoke-DscResource, we don't presently need to
   generate / apply mofs by hand.

   Therefore, remove all templates / classes / related test code.

(maint) remove *-DscConfiguration based templates …

 - Prior to using Invoke-DscResource, there were templates that used
   Start-DscConfiguration and Test-DscConfiguration.

   Remove them as they are unused.


ferventcoder:
LGTM. All good.

PR #43, opened 2015-07-30
(maint) More dead code cleanup / removal

Iristyle:
@jpogran / @ferventcoder some more cleanup action here.  Pretty straightforward stuff.

We'll have to remove the string monkey patch in a separate PR.


lib/puppet/provider/base_dsc/powershell.rb
```diff
@@ -59,7 +82,7 @@ def format_dsc_value(dsc_value)
   def ps_script_content(mode)
     dsc_invoke_method = mode
     @param_hash = resource
-    template = ERB.new(File.new(template_path + "/invoke_dsc_resource.erb").read, nil, '-')
+    template = ERB.new(File.new(template_path + "/invoke_dsc_resource.ps1.erb").read, nil, '-')
```
ferventcoder: Cosmetic change on the template rename since it is all executed in memory. :)

ferventcoder:
As far as the string monkey patching - there is a well-known boolean munging process already existing. I learned it from @hunner and used it in acl.

lib/puppet/provider/base_dsc/powershell.rb
Iristyle: Yup

Iristyle:
@ferventcoder I assume you are talking about https://github.com/puppetlabs/puppetlabs-acl/blob/master/lib/puppet/type/acl.rb#L471-L480

I was thinking we'd do something similar - except take the code from `string.rb`, move it in to the `base_dsc` type, and change the template that handles booleans to call the helper during boolean munging instead of `to_bool` at https://github.com/puppetlabs/puppetlabs-dsc/blob/master/build/dsc/templates/dsc_type.rb.erb#L125-L128

ferventcoder:
> I was thinking we'd do something similar - except take the code from string.rb, move it in to the base_dsc type, and change the template that handles booleans to call the helper during boolean munging instead of to_bool at https://github.com/puppetlabs/puppetlabs-dsc/blob/master/build/dsc/templates/dsc_type.rb.erb#L125-L128

> It's just that that requires regenerating the types again, and I was thinking I'd wait on that bit until after we updated the types as part of #22

And actually, since we have taken the dependency on stdlib, it has a function built in for verifying boolean, it may even have a function for munging boolean. @cyberious @hunner ?

PR #45, opened 2015-07-30
(DO NOT MERGE)(MODULES-2292) Modify Invoke-DscResource Return Values

jpogran:
Modify return values for Invoke-DscResource

### DIFF
--- a/lib/puppet/provider/base_dsc/powershell.rb
+++ b/lib/puppet/provider/base_dsc/powershell.rb
@@ -1,3 +1,5 @@
+require 'json'
+
 Puppet::Type.type(:base_dsc).provide(:powershell) do
   confine :operatingsystem => :windows
   defaultfor :operatingsystem => :windows
@@ -37,29 +39,34 @@ EOT
     script_content = ps_script_content('test')
     Puppet.debug "\n" + script_content
     output = powershell(powershell_args, script_content)
-    if ['true','false'].include?(output.to_s.strip.downcase)
-      check = (output.to_s.strip.downcase == 'true')
-      Puppet.debug "Dsc Resource Exists?: #{check}"
-      Puppet.debug "dsc_ensure: #{resource[:dsc_ensure]}" if resource.parameters.has_key?(:dsc_ensure)
-      Puppet.debug "ensure: #{resource[:ensure]}"
-      check
-    else
-      fail(output)
-    end
+    Puppet.debug "Dsc Resource returned: #{output}"
+    data = JSON.parse(output)
+    fail(data['errormessage']) if !data['errormessage'].empty?
+    check = (data['isdesiredstate'] == 'true')
+    Puppet.debug "Dsc Resource Exists?: #{check}"
+    Puppet.debug "dsc_ensure: #{resource[:dsc_ensure]}" if resource.parameters.has_key?(:dsc_ensure)
+    Puppet.debug "ensure: #{resource[:ensure]}"
+    check
   end
 
   def create
     script_content = ps_script_content('set')
     Puppet.debug "\n" + script_content
     output = powershell(powershell_args, script_content)
-    Puppet.debug "Dsc Resource Return: #{output}"
+    Puppet.debug "Create Dsc Resource returned: #{output}"
+    data = JSON.parse(output)
+    fail(data['errormessage']) if !data['errormessage'].empty?
+    true
   end
 
   def destroy
     script_content = ps_script_content('set')
     Puppet.debug "\n" + script_content
     output = powershell(powershell_args, script_content)
-    Puppet.debug "Dsc Resource Return: #{output}"
+    Puppet.debug "Destroy Dsc Resource returned: #{output}"
+    data = JSON.parse(output)
+    fail(data['errormessage']) if !data['errormessage'].empty?
+    true
   end
 
   def munge_boolean(value)
--- a/lib/puppet/provider/templates/invoke_dsc_resource.ps1.erb
+++ b/lib/puppet/provider/templates/invoke_dsc_resource.ps1.erb
@@ -4,10 +4,17 @@ if (!(Test-Path($PuppetModulesFolder))) {
   & cmd.exe /c mklink /d "$PuppetModulesFolder" "<%= vendored_modules_path %>" | Out-Null
 }
 
+$response = @{
+  indesiredstate = $false
+  rebootrequired = $false
+  errormessage   = ''
+}
+
 $currentState = Get-DscLocalConfigurationManager
 
 if ($currentState.RefreshMode -ne 'Disabled') {
-  throw "DSC LCM RefreshMode must be set to Disabled for Puppet to execute DSC Resources!"
+  $response.errormessage = "DSC LCM RefreshMode must be set to Disabled for Puppet to execute DSC Resources! Please run dsc::lcm_config first"
+  return ($response | ConvertTo-Json -Compress)
 }
 
 $invokeParams = @{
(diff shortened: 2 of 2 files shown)


lib/puppet/provider/base_dsc/powershell.rb
```diff
@@ -52,13 +52,22 @@ def create
     script_content = ps_script_content('set')
     Puppet.debug "\n" + script_content
     output = powershell(powershell_args, script_content)
-    Puppet.debug "Dsc Resource Return: #{output}"
+    if ['true','false'].include?(output.to_s.strip.downcase)
+      if output.to_s.strip.downcase == 'true'
+        # reboot is required
```
jpogran: If a DSC Resource indicates the system needs a reboot, we should tell puppet to do so in both create and destroy. How do we do this?

ferventcoder: That's a good question that should be a discussion item.

Iristyle: New ticket ;0

lib/puppet/provider/templates/invoke_dsc_resource.ps1.erb
```diff
@@ -43,6 +43,6 @@ switch ($invokeParams.Method) {
     [bool]($result.InDesiredState)
   }
   'Set' {
-    @($Null, 0) -contains $result.ReturnValue
+    [bool]($result.RebootRequired)
```
Iristyle: This seems like a peculiar value to return as the succeeds / fails for the `Set` operation.

As I understand it, if `Invoke-DscResource` didn't throw an exception, then the operation succeeded.

Since we don't have the ability to (yet) accept a more structured response in Ruby, I wonder if we're best off just returning `true` here for now.

Otherwise, we have to inspect exit code for pass / fail state, then realize that a `true` means succeeded + reboot required and `false` means succeeded + no reboot required.  Those are some weird semantics.

Alternatively, I think that it would make sense to do the exception catching / structured JSON output now - at least for the results of these operations.  I think that's likely going to take care of  [MODULES-2194](https://tickets.puppetlabs.com/browse/MODULES-2194)  as well which is on this sprints list of things to do.

lib/puppet/provider/templates/invoke_dsc_resource.ps1.erb
```diff
@@ -31,13 +34,25 @@ $invokeParams = @{
...
 }
 
-$result = Invoke-DscResource @invokeParams
+try{
+    $result = Invoke-DscResource @invokeParams
+}catch{
+  $exc = $_
+  $data = [PSCustomObject]@{
+    "error" = "execfailure"
+    "exception" = $_.Exception.Message
```
Iristyle: Message is probably sufficient ... but I have a helper I've written in the past for better parsing of the exception object (SQL stuff won't be relevant of course) - https://github.com/Iristyle/Midori/blob/master/tools/Powershell-Contrib.psm1#L270-L392

jpogran: I didn't know how far we wanted to go with parsing the exception, as line numbers and context won't be as applicable since it's auto-generated in memory

lib/puppet/provider/base_dsc/powershell.rb
```diff
@@ -37,8 +39,13 @@ def exists?
     script_content = ps_script_content('test')
     Puppet.debug "\n" + script_content
     output = powershell(powershell_args, script_content)
-    if ['true','false'].include?(output.to_s.strip.downcase)
-      check = (output.to_s.strip.downcase == 'true')
+    Puppet.debug "Dsc Resource returned: #{output}"
+    data = JSON.parse(output)
+    if data.has_key?("errormessage")
```
jpogran: These if statements could probably be done better

lib/puppet/provider/templates/invoke_dsc_resource.ps1.erb
```diff
@@ -7,7 +7,10 @@ if (!(Test-Path($PuppetModulesFolder))) {
 $currentState = Get-DscLocalConfigurationManager
 
 if ($currentState.RefreshMode -ne 'Disabled') {
-  throw "DSC LCM RefreshMode must be set to Disabled for Puppet to execute DSC Resources!"
+  $data = [PSCustomObject]@{
```
Iristyle: Are there older versions of PowerShell that don't know how to properly serialize a `Hashtable` to JSON or something?

In PS5, this isn't a problem AFAIK

```powershell
@{ "foo" = "bar"} | ConvertTo-Json -Compress


{"foo":"bar"}
```

lib/puppet/provider/templates/invoke_dsc_resource.ps1.erb
```diff
@@ -4,10 +4,17 @@ if (!(Test-Path($PuppetModulesFolder))) {
   & cmd.exe /c mklink /d "$PuppetModulesFolder" "<%= vendored_modules_path %>" | Out-Null
 }
 
+$respone = @{
```
Iristyle: Oops - think you mean `response` and not `respone`

lib/puppet/provider/base_dsc/powershell.rb
```diff
@@ -37,8 +39,11 @@ def exists?
     script_content = ps_script_content('test')
     Puppet.debug "\n" + script_content
     output = powershell(powershell_args, script_content)
-    if ['true','false'].include?(output.to_s.strip.downcase)
-      check = (output.to_s.strip.downcase == 'true')
+    Puppet.debug "Dsc Resource returned: #{output}"
+    data = JSON.parse(output)
+    fail(data['errormessage']) if data['errormessage']
+    if data.has_key?('indesiredstate')
```
Iristyle: The `indesiredstate` key should now always be present

lib/puppet/provider/base_dsc/powershell.rb
```diff
@@ -37,8 +39,11 @@ def exists?
     script_content = ps_script_content('test')
     Puppet.debug "\n" + script_content
     output = powershell(powershell_args, script_content)
-    if ['true','false'].include?(output.to_s.strip.downcase)
-      check = (output.to_s.strip.downcase == 'true')
+    Puppet.debug "Dsc Resource returned: #{output}"
+    data = JSON.parse(output)
+    fail(data['errormessage']) if data['errormessage']
```
Iristyle: If `errormessage` is always present as an empty string, I believe this will always evaluate to true.

That's why I think we likely need another value in the serialized info to track error state of the operation that is set to `true` or `false`

Iristyle: Or alternatively you could use `if !data['errormessage'].empty?`

lib/puppet/provider/templates/invoke_dsc_resource.ps1.erb
```diff
@@ -4,10 +4,17 @@ if (!(Test-Path($PuppetModulesFolder))) {
...
+  indesiredstate = $false
+  rebootrequired = $false
+  errormessage   = ''
+}
+
 $currentState = Get-DscLocalConfigurationManager
 
 if ($currentState.RefreshMode -ne 'Disabled') {
-  throw "DSC LCM RefreshMode must be set to Disabled for Puppet to execute DSC Resources!"
+  $data.errormessage = "DSC LCM RefreshMode must be set to Disabled for Puppet to execute DSC Resources! Please run dsc::lcm_config first"
```
Iristyle: I think you mean `$response.errormessage` and not `$data.errormesage`

lib/puppet/provider/base_dsc/powershell.rb
```diff
@@ -37,29 +39,34 @@ def exists?
...
-      Puppet.debug "dsc_ensure: #{resource[:dsc_ensure]}" if resource.parameters.has_key?(:dsc_ensure)
-      Puppet.debug "ensure: #{resource[:ensure]}"
-      check
-    else
-      fail(output)
-    end
+    Puppet.debug "Dsc Resource returned: #{output}"
+    data = JSON.parse(output)
+    fail(data['errormessage']) if !data['errormessage'].empty?
+    check = (data['isdesiredstate'] == 'true')
```
Iristyle: Oops.. one more typo, but I got you covered.

`isdesiredstate` should be `indesiredstate`

Iristyle:
Tested locally and merged after verification

PR #46, opened 2015-07-30
(maint) don't monkey patch bool munging

Iristyle:
- Create a new munge_boolean helper inside the provider based on the
   current monkey patching.
 - Similarly in the MOF parsing and building step, don't monkey-patch
   String, but instead use a method defined in the template when
   generating tests.

### DIFF
--- a/lib/puppet/provider/base_dsc/powershell.rb
+++ b/lib/puppet/provider/base_dsc/powershell.rb
@@ -62,6 +62,13 @@ EOT
     Puppet.debug "Dsc Resource Return: #{output}"
   end
 
+  def munge_boolean(value)
+    return true if value =~ (/^(true|t|yes|y|1)$/i)
+    return false if value.empty? || value =~ (/^(false|f|no|n|0)$/i)
+
+    raise ArgumentError.new "invalid value: #{value}"
+  end
+
   def format_dsc_value(dsc_value)
     case
     when dsc_value.class.name == 'String'
--- a/lib/puppet/type/base_dsc.rb
+++ b/lib/puppet/type/base_dsc.rb
@@ -1,10 +1,7 @@
 # This is the base type for dsc.
 # Used to inherit the providers for the generated DSC resources
-require 'puppet_x/string'
-
 Puppet::Type.newtype(:base_dsc) do
 
   newparam(:name, :namevar => true ) do
   end
-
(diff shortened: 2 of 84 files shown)


Iristyle:
@ferventcoder / @jpogran I decided to quickly do the string monkey patch removal as well since the impact to the generated types was more minimal than I first anticipated.

lib/puppet/provider/base_dsc/powershell.rb
```diff
@@ -62,6 +62,13 @@ def destroy
     Puppet.debug "Dsc Resource Return: #{output}"
   end
 
+  def munge_boolean(value)
+    return true if value =~ (/^(true|t|yes|y|1)$/i)
+    return false if value.empty? || value =~ (/^(false|f|no|n|0)$/i)
+
+    raise ArgumentError.new "invalid value: #{value}"
+  end
```
ferventcoder: I would have expected the munge boolean in the base type and not the provider.

lib/puppet/type/base_dsc.rb
```diff
@@ -1,10 +1,7 @@
 # This is the base type for dsc.
 # Used to inherit the providers for the generated DSC resources
-require 'puppet_x/string'
-
 Puppet::Type.newtype(:base_dsc) do
 
   newparam(:name, :namevar => true ) do
   end
-
```
ferventcoder: Any reason this wasn't in the base type?

Iristyle: Yeah, it's a good question, and what I originally started to do.  But I couldn't get that quite working given the way the types are defined:

https://github.com/puppetlabs/puppetlabs-dsc/blob/master/lib/puppet/type/dsc_file.rb#L6
https://github.com/puppetlabs/puppetlabs-dsc/blob/master/lib/puppet/type/dsc_file.rb#L8

The `munge_boolean` wasn't actually in the scope of the type.

ferventcoder: I see. Makes sense then.

ferventcoder: Can you add that reasoning as part of the commit message? :)

ferventcoder: Thanks!

PR #48, opened 2015-08-04
[DO NOT MERGE](MODULES-2324) Rake Tasks for Testing

cowofevil:
Add rake tasks for acceptance and integration testing. The Gemfile was
updated to group gems related to installing Puppet into the "development"
group. The Rakefile was modified to put in logic to prevent load errors
if someone didn't install all the Gemfile groups.


cowofevil:
@Iristyle @ferventcoder  @jpogran  I touched a bunch of stuff so you should take a look at the changes.

Gemfile
```diff
@@ -25,91 +25,93 @@ def location_for(place_or_version, fake_version = nil)
   end
 end
 
+# Gems needed by other groups
+gem 'rake', :require => false
+
 group :development do
-  gem 'rake',                         :require => false
```
ferventcoder: It appears this is the only change? 

Usually development is installed with all groups. What cases are you running into that you want to exclude the development group?

ferventcoder:
@cowofevil It appears you backed a few changes away from gemfile? The GemFile is under modulesync, so any changes you make will be switched back by the next sync.

Gemfile
ferventcoder: Technically this change is probably correct. However it seems the whole file was changed.

Rakefile
```diff
@@ -1,22 +1,77 @@
...
-Bundler.require(:rake)
+# build
+if Bundler.rubygems.find_name('mof').any?
+  Dir.glob('build/*.rb').each { |r| import r }
+  Dir.glob('build/*.rake').each { |r| import r }
+end
+
+# tasks
+desc 'Run the Acceptance Tests'
+task :acceptance_tests, :tests, :config, :puppet_agent_ver do |t, args|
```
ferventcoder: I'm under the assumption this is what we normally call?

cowofevil:
@ferventcoder I forgot that the module Gemfile is configured by module-sync. The issue is that I'm running on Windows with Cygwin so the install of Puppet, Facter and Heira is never going to work. Also, for the "acceptance_tests" task and the "integration_tests" tasks none of the development gems are necessary. The only thing that is required is rake to run the tasks.

Rakefile
cowofevil: @ferventcoder This was a bit of experimentation just to see how we may want to call tests from rake tasks in the future. Before we formalize on a rake task testing strategy the test tiering RFC needs to be finalized. (Dom is working on that currently.)

Gemfile
cowofevil: I added the rake gem to the default group because the "system_tests" group only requires rake and no other gems that are part of the "development" group.

cowofevil:
@ferventcoder @Iristyle @jpogran @zreichert Nevermind, I'm closing this PR because it is too ambitious for the way things are currently. Will revisit soon.

Rakefile
ferventcoder: Right on. :)

PR #49, opened 2015-08-05
(MODULES-2175) Create vendored symlink with Ruby

ferventcoder:
Symlinks created the first time during a PowerShell operation don't
always allow the symlink to be used immediately by PowerShell DSC
during the same run.

Prefer instead to create the symlink natively in Ruby. This has
the advantage that PowerShell modules brought in by the symlink
should be ready to be used by the time DSC runs.

Prefer to create the link only once. If the symlink already
exists but points to the wrong location, or if a file/folder
exists where the symlink should be, ensure that the symlink is
recreated to the proper location.

When determining the path for the symlink, check the registry
without redirection to determine where the Program Files
directory is located. `C:\Program Files\WindowsPowerShell\Modules`
is the preferred location for Machine-level PowerShell Modules and
the location where other tools, e.g. PowerShell-Get, also put
modules during installation.


lib/puppet/provider/base_dsc/powershell.rb
```diff
@@ -15,6 +17,18 @@
 Applies DSC Resources by generating a configuration file and applying it.
 EOT
 
+  def initialize(value={})
+    super(value)
+    set_symlink_folder
+  end
+
+  def set_symlink_folder
+    puppetModulesVendorFolder = "#{ENV['ProgramFiles']}\\WindowsPowerShell\\Modules\\PuppetVendoredModules"
```
Iristyle: So a question about this that maybe I should have raised before in the old version... we're making a little bit of an assumption here about location (especially since we're not even looking at `PSModulePath`)

* Do we care about which module path we use since there are a few?  I would think that we want to prefer `Machine` level settings to `User`?
* Are we sure that that module path is always injected for all users (i.e. for `LocalSystem`)?
* Do we want to insulate ourselves from env vars by using the registry (or calling `GetEnvironmentVariable`)?

I did a quick test with my local Administrator:

```
> $env:PSModulePath -split ';'

C:\Users\Administrator\Documents\WindowsPowerShell\Modules
C:\Program Files\WindowsPowerShell\Modules
C:\Windows\system32\WindowsPowerShell\v1.0\Modules\
C:\Program Files (x86)\Microsoft SDKs\Windows Azure\PowerShell\Azure
```

And also with `SYSTEM` (that shows we still get the `Program Files`):

```
PS C:\Windows\system32> $env:PSModulePath -split ';'

WindowsPowerShell\Modules
C:\Program Files\WindowsPowerShell\Modules
C:\Windows\system32\WindowsPowerShell\v1.0\Modules\
C:\Program Files (x86)\Microsoft SDKs\Windows Azure\PowerShell\Azure
```

But note that the `C:\Program Files\WindowsPowerShell\Modules` isn't a configurable `PSModulePath`, so must otherwise be injected.

```ruby
[6] pry(main)> value = ''
[7] pry(main)> Win32::Registry::HKEY_LOCAL_MACHINE.open('SYSTEM\CurrentControlSet\Control\Session Manager\Environment') do |reg|t\Control\Session Manager\Environment') do |reg|

[7] pry(main)*   value = reg['PSModulePath']
[7] pry(main)* end
=> "C:\\Windows\\system32\\WindowsPowerShell\\v1.0\\Modules\\;C:\\Program Files (x86)\\Microsoft SDKs\\Windows Azure\\PowerShell\\Azure"
```

Also, note there are no config values under `HKCU\Environment` for `PSModulePath` either.


So I think that what you're doing here is right (but I might change how we get the path to `ProgramFiles`).

Thoughts?

lib/puppet/provider/base_dsc/powershell.rb
```diff
@@ -2,6 +2,8 @@
   confine :operatingsystem => :windows
   defaultfor :operatingsystem => :windows
 
+  require 'lib/puppet/file_system'
```
Iristyle: nit - I think we usually keep these at the top of the file (even when the provider is confined) - see https://github.com/puppetlabs/puppet/blob/master/lib/puppet/provider/package/windows.rb for example

ferventcoder: Yep. Just setting it first where I know there isn't an issue and will be moving it once things are rolling

lib/puppet/provider/base_dsc/powershell.rb
ferventcoder: Agreed on changing how we get ProgramFiles - in the old way we didn't have the accidental possibility of ending up on `C:\Program Files (x86)` but here we do. Just moving things in at the moment. Working on specs to start defining exactly how this will look.

ferventcoder: We were preferring Program Files as that is still machine but considered where you install modules. It's the same place that PowerShell-Get puts modules. Once I'm ready for this to be reviewed more I will send a note.

lib/puppet/provider/base_dsc/powershell.rb
```diff
@@ -26,7 +66,7 @@ def template_path
   end
 
   def vendored_modules_path
-    File.expand_path(File.join(__FILE__, '..', '..', '..', '..', '..', 'vendor','DSCResources')).gsub(/\//,'\\')
+    "#{@top_level_path}\\vendor\\DSCResources"
```
Iristyle: FYI this is going away pretty soon as part of #53 since files are not pluginsync'd from this directory.

ferventcoder: I will switch that back then.

lib/puppet/provider/base_dsc/powershell.rb
```diff
@@ -15,6 +17,49 @@
 Applies DSC Resources by generating a configuration file and applying it.
 EOT
 
+  def initialize(value={})
+    super(value)
+    set_symlink_folder
+  end
+
+  if Puppet::Util::Platform.windows?
```
ferventcoder: Originally I set this around the method call, but moved it out once I needed access to the registry.

ferventcoder: Which creates the empty method below. Open to a better option

lib/puppet/provider/base_dsc/powershell.rb
```diff
@@ -15,6 +17,49 @@
...
+    def set_symlink_folder
+      #TODO:
+      # - Check whether the path contains a symlink #Puppet::FileSystem.symlink?(puppet_modules_vendor_folder)
+      # - To support v3.4.0, not all of these methods existed with Puppet::FileSystem
+      #   Everything was available in 3.5.0 though based
+      #   on https://github.com/puppetlabs/puppet/commit/938b4e7d495f5f15557110040458171e832cecf2
+      puppet_modules_vendor_folder = "#{get_program_files_dir}\\WindowsPowerShell\\Modules\\PuppetVendoredModules"
+      if !Puppet::FileSystem.exist?(puppet_modules_vendor_folder)
+        Puppet::FileSystem.dir_mkpath(puppet_modules_vendor_folder) if !Puppet::FileSystem.dir_exist?(puppet_modules_vendor_folder)
+        Puppet::FileSystem.symlink(vendored_modules_path, puppet_modules_vendor_folder)
```
ferventcoder: @Iristyle @jpogran @cyberious  Note my comments here, I think we may want to explore the dependency starting at 3.5.x instead of 3.4.0. Symlinking was available in 3.4.x but not all of these methods were available. It may take a little version specific method calls to ensure it is done correctly if we want to support 3.4.0. I haven't looked too deeply into this yet. Writing specs currently.

lib/puppet/provider/base_dsc/powershell.rb
```diff
@@ -15,6 +17,49 @@
...
+    # is in place prior to creating the symlink if the folder doesn't exist
+    def set_symlink_folder
+      #TODO:
+      # - Check whether the path contains a symlink #Puppet::FileSystem.symlink?(puppet_modules_vendor_folder)
+      # - To support v3.4.0, not all of these methods existed with Puppet::FileSystem
+      #   Everything was available in 3.5.0 though based
+      #   on https://github.com/puppetlabs/puppet/commit/938b4e7d495f5f15557110040458171e832cecf2
+      puppet_modules_vendor_folder = "#{get_program_files_dir}\\WindowsPowerShell\\Modules\\PuppetVendoredModules"
+      if !Puppet::FileSystem.exist?(puppet_modules_vendor_folder)
+        Puppet::FileSystem.dir_mkpath(puppet_modules_vendor_folder) if !Puppet::FileSystem.dir_exist?(puppet_modules_vendor_folder)
```
ferventcoder: The top level folder should already exist in almost all scenarios, except when testing on a machine that doesn't have WMF 4+ installed. This extra check ensures the folder structure is in place.

ferventcoder: I think unit tests are going to need to stub this and the next line.

lib/puppet/provider/base_dsc/powershell.rb
```diff
@@ -15,6 +17,49 @@
...
+        Puppet::FileSystem.symlink(vendored_modules_path, puppet_modules_vendor_folder)
+      end
+    end
+
+    # get the absolute program files path, no matter whether the process is
+    # a 32 bit or 64 bit process. Shut off registry redirection and query
+    # the ProgramFilesDir key to retrieve the program files value.
+    def get_program_files_dir
+      program_files_dir = 'C:\Program Files'
```
Iristyle: Set you're not using `||= reg['ProgramFilesDir']` below, so I don't think this default is ever used.

Can probably just change line 50 to the following (though `reg['missing_value']` might `throw` anyhow, making that moot)

```ruby
reg['ProgramFilesDir'] || 'C:\\Program Files'
```

lib/puppet/provider/base_dsc/powershell.rb
Iristyle: I think you mean `exist?` / `dir_mkpath` /  `dir_exist?` right?

I'm OK with a `3.5` minimum given our feature page says Puppet 3.8 / Puppet 4.0 as the target Puppet versions.

/cc @cowofevil

lib/puppet/provider/base_dsc/powershell.rb
ferventcoder: I should have started with nil :)

I also believe it will throw if the value doesn't exist.

lib/puppet/provider/base_dsc/powershell.rb
Iristyle: In Puppet there is: https://github.com/puppetlabs/puppet/blob/master/lib/puppet/util/windows.rb#L2-L8 for the sake of tests.  You could do something like this (i.e. define `set_symlink_folder` and then redefine if on Windows).

Or you could do something similar to what was done in sqlserver
https://github.com/puppetlabs/puppetlabs-sqlserver/blob/36a45a89df43f8802a9882d87443bc43ec86d74b/lib/puppet_x/sqlserver/sql_connection.rb

lib/puppet/provider/base_dsc/powershell.rb
ferventcoder: Yes, the specific methods not available are the following:

* `Puppet::FileSystem.exist?` (available if you have an instance though)
* `Puppet::FileSystem.dir_mkpath` 
* `Puppet::FileSystem.dir_exist?` 
* `Puppet::FileSystem.symlink` (@Iristyle this one, too. Requires an instance of Puppet::FileSystem::File)

lib/puppet/provider/base_dsc/powershell.rb
ferventcoder: I think we did something for win32console. The first example was about classes and module definitions, not methods.

ferventcoder: `define_method` seemed a bit much for what I want to do here.

lib/puppet/provider/base_dsc/powershell.rb
ferventcoder: https://github.com/puppetlabs/puppet/blob/3.4.0/lib/puppet/file_system/file.rb#L207-L214 For symlink requiring an instance.

lib/puppet/provider/base_dsc/powershell.rb
ferventcoder: Darn, looks like the methodology for Win/Linux in colors.rb was similar - https://github.com/puppetlabs/puppet/blob/3.8.0/lib/puppet/util/colors.rb#L79-L190

spec/unit/puppet/provider/powershell_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+  end
+
+  it "should be an instance of Puppet::Type::Base_dsc::ProviderPowershell" do
+    provider.must be_an_instance_of Puppet::Type::Base_dsc::ProviderPowershell
+  end
+
+  context ".set_symlink_folder" do
+    context "on Windows", :if => Puppet.features.microsoft_windows? do
+      before :each do
+        #delete the symlink if it exists
```
ferventcoder: This will be likely in the integration spec. Placeholder to remind myself to add.

lib/puppet/provider/base_dsc/powershell.rb
```diff
@@ -34,6 +88,7 @@ def powershell_args
   end
 
   def exists?
+    set_symlink_folder
```
ferventcoder: If exists? is always called for create and destroy then this is a good place to set. Putting `set_symlink_folder` in initialize made it extremely hard to mock out. I think it's still possible if that is the right place to put it, but it will get interesting.

Iristyle: Do we cause problems by not re-symlinking each run?

For instance, if you're developing locally / testing and you happen to move source... by not resymlinking to the current directory, you'll be pointing to the wrong thing.

Not sure if there are any legit runtime scenarios where that would be a problem or not.

ferventcoder: Yep, note the todo information in the method.

spec/unit/puppet/provider/powershell_spec.rb
```diff
@@ -0,0 +1,131 @@
...
+        Puppet::FileSystem.expects(:symlink).never
+
+        provider.set_symlink_folder
+      end
+    end
+  end
+
+  context ".get_program_files_dir", :if => Puppet.features.microsoft_windows? do
+
+    let (:reg_key) { mock('Win32_Registry') }
```
ferventcoder: Can't use this, and I can't use `let (:reg_key) { Win32::Registry}`. Both of them will register correctly, but mocks won't iterate over the evaluation block in a way that actually sets the `program_files_dir` variable.

spec/unit/puppet/provider/powershell_spec.rb
```diff
@@ -0,0 +1,131 @@
...
+
+    before :each do
+      Win32::Registry::HKEY_LOCAL_MACHINE.expects(:open).
+        with('SOFTWARE\Microsoft\Windows\CurrentVersion', Win32::Registry::KEY_READ | 0x100).
+        returns(reg_key)
+    end
```
ferventcoder: This all goes away as well because I'm moving to the nuclear `any_instance` option. It has the downside that the registry query will actually occur, but the upside is that the location will always exist and the specs are only being performed on Windows. Ugh.

metadata.json
```diff
@@ -29,11 +29,11 @@
   "requirements": [
     {
       "name": "pe",
-      "version_requirement": "3.x"
+      "version_requirement": ">= 3.3.0 < 2015.3.0"
     },
     {
       "name": "puppet",
-      "version_requirement": "3.x"
+      "version_requirement": ">= 3.5.0 < 5.0.0"
```
ferventcoder: Bumped to 3.5.0 for the reasons listed in the first commit. First version of Puppet with the FileSystem non-instance calls that are used.

spec/unit/puppet/provider/powershell_spec.rb
```diff
@@ -0,0 +1,126 @@
...
+require 'puppet/type/base_dsc'
+
+describe Puppet::Type.type(:base_dsc).provider(:powershell) do
+
+  # Override initialize so we don't call the symlink set
+  class Puppet::Type::Base_dsc::ProviderPowershell
+    def initialize(value={})
+      super(value)
+    end
+  end
```
ferventcoder: I'm hoping this doesn't have an effect on the regular use of Puppet since this is in the specs.

ferventcoder:
Everything passes for specs. I'm curious to know if there are actual issues in a run though. Going to test this.

lib/puppet/provider/base_dsc/powershell.rb
```diff
@@ -15,6 +17,61 @@
...
+          recreate_link = true
+        else
+          recreate_link = true if Puppet::FileSystem.readlink(puppet_modules_vendor_folder) != vendored_modules_path
+          Puppet::FileSystem.unlink(puppet_modules_vendor_folder) if recreate_link
+        end
+      end
+      if recreate_link || !Puppet::FileSystem.exist?(puppet_modules_vendor_folder) || !Puppet::FileSystem.symlink?(puppet_modules_vendor_folder)
+        Puppet.debug "Creating symlink at '#{puppet_modules_vendor_folder}' \n pointed to '#{vendored_modules_path}'."
+        Puppet::FileSystem.dir_mkpath(puppet_modules_vendor_folder) if !Puppet::FileSystem.dir_exist?(puppet_modules_vendor_folder)
+        Puppet::FileSystem.symlink(vendored_modules_path, puppet_modules_vendor_folder,{:force => true})
```
ferventcoder: It doesn't appear that `:force => true` actually works. @Iristyle am I not passing this right?

ferventcoder:
@Iristyle Okay, this will pass specs, but it is still not good. Will not override an existing folder with the same name.

lib/puppet/provider/base_dsc/powershell.rb
ferventcoder: I should have read the docs a little better. If it's a directory it doesn't do anything. https://github.com/puppetlabs/puppet/blob/master/lib/puppet/file_system.rb#L230-L235

ferventcoder: BTW, the behavior being inconsistent between file/directory is :-1:

spec/unit/puppet/provider/powershell_spec.rb
ferventcoder: It doesn't. All good.

spec/spec_helper.rb
```diff
@@ -14,7 +15,31 @@
...
+  oldtmpdir = Dir.tmpdir()
+  ENV['TMPDIR'] = tmpdir
+
+  c.after :suite do
+    # return to original tmpdir
+    ENV['TMPDIR'] = oldtmpdir
+    if Puppet::Util::Platform.windows?
+      #FileUtils.rm_rf(tmpdir) - follows symlinks on Windows
+      # and deletes there as well :/
+      system("rmdir /s /q \"#{tmpdir}\"")
```
ferventcoder: This looks like a bug in the FileUtils.rm_rf - it will follow the symlinks and delete the files. :/

I don't know if this is expected behavior, so I shared on [Stack Overflow](http://stackoverflow.com/a/609060/18475) and on [twitter](https://twitter.com/ferventcoder/status/629667387458088960)

ferventcoder: Either way, this may need a pure verification. The symlinks are created calling `Puppet::FileSystem.symlink`, which calls `CreateSymbolicLinkW` - https://github.com/puppetlabs/puppet/blob/b3d9c03004bcb09d2539d48db666a2921f9bc481/lib/puppet/util/windows/file.rb#L82-L90

spec/integration/puppet/provider/powershell_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+  let (:path) { top_level_path }
+  let (:program_files) { "#{path}\\Program Files"}
+  let (:vendored_modules_symlink_path){ "#{program_files}\\WindowsPowerShell\\Modules\\PuppetVendoredModules" }
+
+  context ".set_symlink_folder" do
+    before :each do
+      Win32::Registry.any_instance.expects(:[]).with('ProgramFilesDir').returns(program_files).at_most(10)
+      Puppet::FileSystem.unlink(vendored_modules_symlink_path) if Puppet::FileSystem.symlink?(vendored_modules_symlink_path)
+      system("rmdir /s /q \"#{program_files}\"") if ::File.directory?(program_files)
+      #FileUtils.rm_r(program_files) if ::File.directory?(program_files)
```
ferventcoder: This looks like a bug in the FileUtils.rm_rf - it will follow the symlinks and delete the files. :/

I don't know if this is expected behavior, so I shared on [Stack Overflow](http://stackoverflow.com/a/609060/18475) and on [twitter](https://twitter.com/ferventcoder/status/629667387458088960)

ferventcoder: Either way, this may need a pure verification. The symlinks are created calling `Puppet::FileSystem.symlink`, which calls `CreateSymbolicLinkW` - https://github.com/puppetlabs/puppet/blob/b3d9c03004bcb09d2539d48db666a2921f9bc481/lib/puppet/util/windows/file.rb#L82-L90

ferventcoder: Note that an unlink is done here, but at the end of the suite run we also call a cleanup that was also calling `FileUtils.rm_rf` - see https://github.com/puppetlabs/puppetlabs-dsc/pull/49/files#diff-93830fa29d616f7c87903d08b5b1b29aR30

ferventcoder: @Iristyle I think `Puppet::FileSystem` needs some enhancements for file management (create/delete/move). Thoughts?

lib/puppet/provider/base_dsc/powershell.rb
```diff
@@ -15,6 +18,62 @@
...
+      recreate_link = false
+      if Puppet::FileSystem.exist?(puppet_modules_vendor_folder) || Puppet::FileSystem.symlink?(puppet_modules_vendor_folder)
+        if !Puppet::FileSystem.symlink?(puppet_modules_vendor_folder)
+          recreate_link = true
+          FileUtils.mv puppet_modules_vendor_folder, "#{puppet_modules_vendor_folder}.#{Time.now.utc.strftime("%Y%m%d_%H%M%S_%L")}"
+        else
+          recreate_link = true if Puppet::FileSystem.readlink(puppet_modules_vendor_folder) != vendored_modules_path
+          Puppet::FileSystem.unlink(puppet_modules_vendor_folder) if recreate_link
+        end
+      end
```
ferventcoder: I'm going to refactor this a bit.

spec/integration/puppet/provider/powershell_spec.rb
cyberious: I would also suggest removing the comment line.

ferventcoder: Will do. Wanted it here for discussion.

lib/puppet/provider/base_dsc/powershell.rb
cyberious: Any thought of breaking this out into a separate helper file to keep the type simpler to read.

spec/integration/puppet/provider/powershell_spec.rb
ferventcoder: And cleaned up.

lib/puppet/provider/base_dsc/powershell.rb
ferventcoder: We've already combined helper files into just this one provider file.

lib/puppet/provider/base_dsc/powershell.rb
ferventcoder: Done. Should be simpler to understand now.

lib/puppet/provider/base_dsc/powershell.rb
```diff
@@ -15,6 +18,66 @@
...
+    # Ensure the symlink folder exists and is pointing to the proper location.
+    # This also ensures the directory structure is in place prior to creating
+    # the symlink, otherwise the symlink creation fails.
+    def set_symlink_folder
+      puppet_modules_vendor_folder = "#{get_program_files_dir}\\WindowsPowerShell\\Modules\\PuppetVendoredModules"
+      recreate_link = false
+
+      if Puppet::FileSystem.exist?(puppet_modules_vendor_folder)
+        if Puppet::FileSystem.symlink?(puppet_modules_vendor_folder)
+          return if Puppet::FileSystem.readlink(puppet_modules_vendor_folder) == vendored_modules_path
```
ferventcoder: Jumping out early here. Trying to reduce the additional symlink call if we've determined we are all good. This will be the normal code path 99% of the time so I tried to optimize a bit for it.

ferventcoder: I could take it a step further and move it up to line 34 and check only `symlink?` and `readlink` path and return. That may be clearer.

ferventcoder: Bumped it up to the top of the method.

lib/puppet/provider/base_dsc/powershell.rb
```diff
@@ -15,6 +18,67 @@
...
+  if Puppet::Util::Platform.windows?
+    require 'win32/registry'
+
+    # Ensure the symlink folder exists and is pointing to the proper location.
+    # This also ensures the directory structure is in place prior to creating
+    # the symlink, otherwise the symlink creation fails.
+    def set_symlink_folder
+      puppet_modules_vendor_folder = "#{get_program_files_dir}\\WindowsPowerShell\\Modules\\PuppetVendoredModules"
+      return if Puppet::FileSystem.symlink?(puppet_modules_vendor_folder) &&
+                Puppet::FileSystem.readlink(puppet_modules_vendor_folder) == vendored_modules_path
```
ferventcoder: Return early if we have our expected symlink.

appveyor.yml
```diff
@@ -33,7 +33,7 @@ build: off
 test_script:
 - bundle exec puppet -V
 - ruby -v
-- bundle exec rspec spec/unit -fd -b
+- bundle exec rspec spec/unit spec/integration -fd -b
```
ferventcoder: Ensure integration specs are run

metadata.json
Iristyle: :+1:

lib/puppet/provider/base_dsc/powershell.rb
Iristyle: As far as order of operations here... 

I think we want the symlink creation only once when the provider is first created, not each time a new provider instance is associated with a path though, right?  Which I think means we want this in `self.prefetch` instead of in the initializer.

I'm testing this locally now (and will make sure I understand the testing implications, given I know you worked hard on getting all that setup).

Iristyle:
Still a WIP, but you can keep an eye on proposed changes at https://github.com/ferventcoder/puppetlabs-dsc/compare/ticket/master/MODULES-2175_symlinks...Iristyle:pr/49?expand=1

spec/unit/puppet/provider/powershell_spec.rb
```diff
@@ -0,0 +1,157 @@
...
+
+  let (:resource) { Puppet::Type.type(:base_dsc).new(:provider => :powershell, :name => "windows_provider") }
+  let (:provider) { resource.provider }
+
+  before :each do
+    resource.provider = provider
```
Iristyle: I don't really understand what this is for?

In my proposed changes I've axed it, since you can just use `subject` everywhere (and it's a different object instance within each test which I *think* is what you were looking for?).

ferventcoder: No idea either. I hadn't axed it yet.

Out of curiosity, why are you changing what is here?

ferventcoder: IOW, if you want me to look at prefetch... you could just state that here.

lib/puppet/provider/base_dsc/powershell.rb
ferventcoder: Are we worried at all about using prefetch for something different than intended?

spec/unit/puppet/provider/powershell_spec.rb
Iristyle: Only because it's easiest for me right now to actually run / modify the code / tests locally - plus I thought you were out for the day ;0

ferventcoder: Oh right on. :) 

I was. I'm thinking further enhancements could be a new pull request. So that we can unblock testing.

lib/puppet/provider/base_dsc/powershell.rb
Iristyle: It definitely seems weird to use `prefetch` for this purpose, I agree... 

I think that the most correct thing is probably to use a feature, which I had kind of forgotten about.  I moved the code over to a feature and made a few tweaks to make it run, but haven't modified tests yet.  I pushed it up to the branch in my repo (note I haven't fixed tests yet):

https://github.com/ferventcoder/puppetlabs-dsc/compare/ticket/master/MODULES-2175_symlinks...Iristyle:pr/49?expand=1

I think a feature is probably the right place for this.  Then we also address the @cyberious point about separating the "vendoring through symlink" code from the provider.  And I believe the feature initialization only runs once as well.

Iristyle:
Ok @ferventcoder , finished moving all the specs over along with the `vendors_dsc?` feature - see latest at https://github.com/ferventcoder/puppetlabs-dsc/compare/ticket/master/MODULES-2175_symlinks...Iristyle:pr/49?expand=1

Interestingly, I made a minor refactor in the final commit - and it now causes an intermittent test failure.  I haven't had the time to track that down yet though - but if you have any ideas, that would be good.

ferventcoder:
@Iristyle specs are running, but everything is squashed down and ready.

ferventcoder:
AppVeyor died on S3 issue. Rekicking.

PR #50, opened 2015-08-05
(MODULES-2336) Create Tests for Basic DSC Resource "Service"

cowofevil:
Add negative tests cases for the "service" DSC resource type.


zreichert:
+1

PR #53, opened 2015-08-05
(MODULES-2322) DSC resources need to pluginsync

Iristyle:
- The vendor directory is not something that will pluginsync from
   a master to agent.  Move everything in vendor/DSCResources to
   lib/puppet_x/dsc_resources instead.

 - As a result, the build/dsc.rake task must be updated to produce
   files in the new directory.
 - Also as a result, the base_dsc/powershell provider needs to find
   the resources in a different directory.

   Note that the relative paths layout was verified with both puppet
   apply locally and with puppet agent after a master to agent
   pluginsync


Iristyle:
@ferventcoder / @cowofevil this is g2g

ferventcoder:
rebase time. Joy... :0

PR #54, opened 2015-08-06
(MODULES-2365) Allow Tests to be Run Locally

cowofevil:
The acceptance and integration tests can only be ran against artifacts
on the module forge. This is problematic for developers that want to test
their changes before commiting code. The changes made update the test run
scripts, configs and documentation to allow for local testing.

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -63,9 +63,14 @@ To run acceptance tests use the "acceptance_tests.sh" test run script.
 ./acceptance.sh
 ```
 
-**Example: Run with Puppet Agent 1.2.1 on Windows 2008 R2**
+**Example: Run with Puppet Agent 1.2.1 on Windows 2008 R2 using Forge**
 ```
-./acceptance.sh windows-2008r2-64a 1.2.1
+./acceptance.sh windows-2008r2-64a 1.2.1 forge
+```
+
+**Example: Run with Puppet Agent 1.2.2 on Windows 2012 R2 with local module code (No Forge)**
+```
+./acceptance.sh windows-2012r2-64a 1.2.2 local
 ```
 
 ### Running Integration Tests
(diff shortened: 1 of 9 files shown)


tests/test_run_scripts/acceptance_tests.sh
```diff
@@ -12,8 +12,9 @@ declare -a ARGS
 if [ $# -eq 0 ]; then
   ARGS[0]='windows-2012r2-64a'
   ARGS[1]='1.2.2'
-elif [ $# -ne 2 ]; then
-  echo 'USAGE acceptance_tests.sh <CONFIG> <PUPPET_AGENT_VERSION>'
+  ARGS[2]='true'
```
Iristyle: Passing true at the command line like this is a little weird... but I suspect you'll be the only one using this script @cowofevil

cowofevil: Would you prefer "forge" and "local" instead?

Iristyle: Yeah, something that indicates the behavior would be a better naming choice IMHO.

Iristyle:
Better @cowofevil  - squash 'er up!

cowofevil:
@Iristyle Squashed!

tests/integration/pre-suite/01_dsc_module_install.rb
```diff
@@ -3,3 +3,16 @@
 step 'Install Module via PMT'
 stub_forge_on(master)
 on(master, puppet('module install puppetlabs-dsc'))
```
Iristyle: This line should have been removed.

tests/README.md
```diff
@@ -77,9 +82,14 @@ To run integration tests use the "integration_tests.sh" test run script.
...
-**Example: Run with alternate PE package repo on Windows 2008 R2**
+**Example: Run with alternate PE package repo on Windows 2008 R2 using Forge**
+```
+./integration.sh windows-2008r2-64mda http://alt.address.local/3.8.2/preview forge
+```
+
+**Example: Run with alternate PE package repo on Windows 2012 R2 with local module code (No Forge)**
 ```
-./integration.sh windows-2008r2-64mda http://alt.address.local/3.8.2/preview
+./acceptance.sh windows-2012r2-64a http://alt.address.local/3.8.2/preview local
```
Iristyle: this should be `integration`, not `acceptance`

Iristyle:
I've fixed up some of the problems / testing locally.  Will merge that through and close this PR.

tests/README.md
Iristyle: And also noticed that these should be `acceptance_tests` and `integration_tests` now as well.  Fixing up in my local testing.

PR #57, opened 2015-08-12
(MODULES-2388) Warning Output breaks JSON Parsing

jpogran:
Invoke-DscResource does not respect WarningAction or WarningVariable and
so sends output to the Warning stream regardlees of user setting. This
Warning output is caught and sent back along with our objects, and
breaks the JSON parser.

We set $WarningPreference to 'SilentlyContinue' to work around this.


cowofevil:
LGTM +1

I ran the following command and no horrid output:

```
puppet apply -e "dsc_xcomputer { 'test': dsc_name => 'catflakes' }"
```

PR #61, opened 2015-08-19
(docs) Add note about long file names during install

cyberious:
- MODULES-2408 describe what issues can arise during install
  due to long file names and the username used during install

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,15 @@ This module is available on the [Puppet Forge](https://forge.puppetlabs.com/pupp
 puppet module install puppetlabs-dsc
 ~~~
 
+**NOTE**
+
+When installing the module on Windows you may run into an issue regarding long file names (LFN), this is due to the generated schema files. If you are running in a master environment and installing the module on your master this will not be an issue, however if you are running masterless then you may run into this during installation depending on the length of the username it is ran from.  One way to get around this is to download the `.tar.gz` and install using the file rather than directly from the forge. For more info see [PUP-4866](https://tickets.puppetlabs.com/browse/PUP-4866)
+
+~~~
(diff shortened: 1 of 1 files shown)


cyberious:
@ferventcoder you want the full link or just a reference to the ticket?

cyberious:
@ferventcoder added in README the link

ferventcoder:
LGTM.

jpogran:
lgtm

README.md
```diff
@@ -23,6 +23,15 @@ This module is available on the [Puppet Forge](https://forge.puppetlabs.com/pupp
 puppet module install puppetlabs-dsc
 ~~~
 
+**NOTE**
+
+When installing the module on Windows you may run into an issue regarding long file names (LFN), this is due to the generated schema files. If you are running in a master environment and installing the module on your master this will not be an issue, however if you are running masterless then you may run into this during installation depending on the length of the username it is ran from.  One way to get around this is to download the `.tar.gz` and install using the file rather than directly from the forge. For more info see [PUP-4866](https://tickets.puppetlabs.com/browse/PUP-4866)
```
ferventcoder: After closer inspection, this isn't quite right. We've provided a fix in https://tickets.puppetlabs.com/browse/PUP-4854 that will switch to using TEMP directory in Puppet 3.8.2 and 4.2.2. That means then you may run into this issue based on username and how they access temp, but before it was to C:\ProgramData\ etc etc. - it may help to review https://github.com/puppetlabs/puppet/pull/4090

cyberious: So would it be safer to say, if using < 3.8.2 in or > 4 and less than 4.2.2 you may get this issue otherwise it has been fixed?

ferventcoder: It hasn't been fixed though. We still don't support LFN, we've just attempted to ease the likelihood of if happening in PUP-4854. 

I think this should actually do into known issues.

ferventcoder:
Closed by #68.

PR #64, opened 2015-08-21
(MODULES-2442) Mixing Quotes Causes Crash

jpogran:
If a user specifies single quotes inside any string in a puppet manifest our module
will produce an invalid quoted string. We can fix this by properly escaping the single quote
inside the string provided.

### DIFF
--- a/lib/puppet/provider/base_dsc/powershell.rb
+++ b/lib/puppet/provider/base_dsc/powershell.rb
@@ -78,7 +78,7 @@ EOT
   def format_dsc_value(dsc_value)
     case
     when dsc_value.class.name == 'String'
-      "'#{dsc_value}'"
+      "'#{escape_quotes(dsc_value)}'"
     when dsc_value.class.ancestors.include?(Numeric)
       "#{dsc_value}"
     when [:true, :false].include?(dsc_value)
@@ -90,6 +90,10 @@ EOT
     else
       fail "unsupported type #{dsc_value.class} of value '#{dsc_value}'"
     end
+  end
+
+  def escape_quotes(text)
+    text.gsub("'", "''")
   end
 
   def ps_script_content(mode)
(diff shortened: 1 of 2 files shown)


ferventcoder:
Do not merge because the specs are not done yet?

jpogran:
Yep. Specs added now.

spec/unit/puppet/provider/powershell_input_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+        expect(@provider.ps_script_content('test')).to match(/contents = 'The ''Cats'' go "meow"!'/)
+      end
+
+      it "should handle dollar signs" do
+        dsc_file[:dsc_contents] = "This should show \$foo variable"
+        expect(@provider.ps_script_content('test')).to match(/contents = 'This should show \$foo variable'/)
+      end
+    end
+  end
+end
```
ferventcoder: I'd put this in the powershell_spec.rb.

spec/unit/puppet/provider/powershell_input_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+    )
+  end
+
+  describe "powershell provider" do
+
+    before(:each) do
+      @provider = example_dsc_resource.provider(:powershell).new(dsc_file)
+    end
+
+    describe "when quotes are present" do
```
ferventcoder: Will need an option where quotes are not present. 

Otherwise mostly good. I'm curious if there is a better way to get to the provider. :)

spec/unit/puppet/provider/powershell_input_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+
+      it "should handle single quotes" do
+        dsc_file[:dsc_contents] = "The 'Cats' go 'meow'!"
+        expect(@provider.ps_script_content('test')).to match(/contents = 'The ''Cats'' go ''meow''!'/)
```
ferventcoder: I'd prefer this call be to `@provider.format_dsc_value` - possibly with an expect on `dsc_value.class.name` that returns `String`.

PR #66, opened 2015-08-24
(FM-2564) Ensure rake task copies DSC resources to vendor folder

jpogran:
A bug in the rake task meant the cloned xDscResources did not copy over to the lib/puppet_x/dsc_resources folder. This fixes it


ferventcoder:
We don't use `(bug)` that I am aware of. Please adjust the commit message to `(FM-2564)` and the rest. Otherwise :+1:

PR #69, opened 2015-08-25
(MODULES-2299) Ensure DSC runs under Puppet 4.2.0

ferventcoder:
The current method of pushing a directory to the loadpath that is used
in the types is considered non-standard and may cause unintended
consequences. Instead of adjusting the load path, require using a
relative location from the current source file directory.

`vendors_dsc?` should not include the question mark in the name. That
can confuse Puppet when it creates the predicate
`Puppet.features.featurename?` where in this case it may end up
creating `Puppet.features.vendors_dsc??`. Rename the feature to
`vendors_dsc` instead.

Additionally, having the symlink module code inside the feature is
confusing Linux Masters, with the failure happening on the early return
false at
https://github.com/puppetlabs/puppetlabs-dsc/blob/662a2961cd4f20e2e11fb82ae4778ca53071e9e5/lib/puppet/feature/vendors_dsc.rb#L7
Instead, move the symlink code out to it's own module and require that
in the feature add.

Paired with James Pogran
Paired with Josh Cooper

### DIFF
--- a/build/dsc/templates/dsc_type.rb.erb
+++ b/build/dsc/templates/dsc_type.rb.erb
@@ -1,9 +1,7 @@
-# workaround for cross modules dependencies
-master_path = File.expand_path(File.join(__FILE__, '..', '..', '..', '..', '..', '<%= @module_path.basename.to_s.split('-').last %>','lib'))
-$:.push(master_path) if File.directory?(master_path)
-require 'puppet/type/base_dsc'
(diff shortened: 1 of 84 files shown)


jpogran:
Verified on win2012, ruby 2.1.5, and puppet 4.2.0. LGTM

Iristyle:
:+1:

PR #74, opened 2015-08-27
(MODULES-2480) Create Negative Test for LCM Configuration

cowofevil:
Add an acceptance test for misconfiguring the LCM before running DSC manifests.
I added a new method to the "dsc_utils" helper library to allow for easy
transition between refresh modes.

### DIFF
--- a/tests/acceptance/tests/basic_functionality/negative/misconfigure_lcm.rb
+++ b/tests/acceptance/tests/basic_functionality/negative/misconfigure_lcm.rb
@@ -9,7 +9,6 @@ local_files_root_path = ENV['MANIFESTS'] || 'tests/manifests'
 
 # ERB Manifest
 dsc_type = 'file'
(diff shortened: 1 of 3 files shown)


tests/acceptance/tests/basic_functionality/negative/misconfigure_lcm.rb
```diff
@@ -0,0 +1,41 @@
...
+test_name 'MODULES-2480 - C89505 - Misconfigure LCM and Attempt to Apply DSC Manifest'
+
+confine(:to, :platform => 'windows')
+
+# Init
+local_files_root_path = ENV['MANIFESTS'] || 'tests/manifests'
+
+# ERB Manifest
+dsc_type = 'file'
+dsc_module = 'PSDesiredStateConfiguration'
```
zreichert: I don't think this variable is used in 'dsc_single_resource.pp.erb'

PR #73, opened 2015-08-27
(DO NOT MERGE)(FM-3578) Update xDscResources from PowerShell Repo

jpogran:
Import all the current xDscResources from the PowerShell Github Repo and
update all xDscResource types

Depends on #76

PR #76, opened 2015-08-28
(DO NOT MERGE)(MODULES-2479) Allow Different Ensure Values

jpogran:
Some xDscResources use different values of Ensure than the standard
present/absent. Add new logic to detect when different values are used
and use them for generating the type, the spec, and the powershell
script.

### DIFF
--- a/build/dsc/resource.rb
+++ b/build/dsc/resource.rb
@@ -82,7 +82,26 @@ module Dsc
 
     def absentable?
       properties.detect do |p|
-        p.is_ensure? && p.values.any? { |v| v.casecmp('absent') == 0 }
+        return 'absent' if p.is_ensure? && p.values.any? { |v| v.casecmp('absent') }
+        return 'disable' if p.is_ensure? && p.values.any? { |v| v.casecmp('disable') }
+      end
+    end
+
+     def absent_value
+      properties.detect do |p|
+        p.is_ensure? && p.values.each do |v|
+          return 'absent' if v.casecmp('absent') == 0
+          return 'disable' if v.casecmp('disable') == 0
+        end
+      end
+    end
+
+    def ensure_value
+      properties.detect do |p|
+        p.is_ensure? && p.values.each do |v|
+          return 'present' if v.casecmp('present') == 0
+          return 'enable' if v.casecmp('enable') == 0
+        end
       end
     end
 
--- a/build/dsc/templates/dsc_type.rb.erb
+++ b/build/dsc/templates/dsc_type.rb.erb
@@ -5,6 +5,19 @@ Puppet::Type.newtype(:dsc_<%= resource.friendlyname.downcase %>) do
 
   provide :powershell, :parent => Puppet::Type.type(:base_dsc).provider(:powershell) do
     defaultfor :operatingsystem => :windows
+
+<% if resource.ensurable? -%>
+    def ensure_value
+        '<%= resource.ensure_value %>'
+    end
+<% end -%>
+
+<% if resource.absentable? -%>
+    def absent_value
+        '<%= resource.absent_value %>'
+    end
+<% end -%>
+
   end
 
   @doc = %q{
@@ -71,7 +84,7 @@ Puppet::Type.newtype(:dsc_<%= resource.friendlyname.downcase %>) do
 <%    end -%>
     validate do |value|
 <%    if property.name.downcase == 'ensure' -%>
-      resource[:ensure] = value.downcase
+      resource[:ensure] = provider.munge_ensure(value)
 <%    end -%>
 <%    case -%>
 <%    when property.array? -%>
--- a/build/dsc/templates/dsc_type_spec.rb.erb
+++ b/build/dsc/templates/dsc_type_spec.rb.erb
@@ -78,7 +78,7 @@ describe Puppet::Type.type(:dsc_<%= resource.friendlyname.downcase %>) do
 
   it 'should accept dsc_ensure predefined value <%= value.downcase %> and update ensure with this value (ensure end value should be a symbol)' do
     dsc_<%= resource.friendlyname.downcase %>[:dsc_ensure] = '<%= value.downcase %>'
-    expect(dsc_<%= resource.friendlyname.downcase %>[:ensure]).to eq(dsc_<%= resource.friendlyname.downcase %>[:dsc_ensure].downcase.to_sym)
+    expect(dsc_<%= resource.friendlyname.downcase %>[:ensure]).to eq(dsc_<%= resource.friendlyname.downcase %>.provider.munge_ensure(dsc_<%= resource.friendlyname.downcase %>[:dsc_ensure].downcase).to_sym)
   end
 <%        end # if property.name.downcase.... -%>
 <%      end # property.values.each....-%>
@@ -203,11 +203,11 @@ describe Puppet::Type.type(:dsc_<%= resource.friendlyname.downcase %>) do
 <%  end # unless ['msft_filedirectoryconfiguration'].... -%>
 <%  if resource.ensurable? -%>
 
-    describe "when dsc_ensure is 'present'" do
+    describe "when dsc_ensure is '<%= resource.ensure_value %>'" do
 
       before(:each) do
-        dsc_<%= resource.friendlyname.downcase %>.original_parameters[:dsc_ensure] = 'present'
-        dsc_<%= resource.friendlyname.downcase %>[:dsc_ensure] = 'present'
+        dsc_<%= resource.friendlyname.downcase %>.original_parameters[:dsc_ensure] = '<%= resource.ensure_value %>'
+        dsc_<%= resource.friendlyname.downcase %>[:dsc_ensure] = '<%= resource.ensure_value %>'
         @provider = described_class.provider(:powershell).new(dsc_<%= resource.friendlyname.downcase %>)
       end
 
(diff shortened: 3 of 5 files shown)


build/dsc/resource.rb
```diff
@@ -82,7 +82,37 @@ def ensurable?
...
+
+    def absent_opposite?
+      properties.detect do |p|
+        p.is_ensure? && p.values.any? { |v| v.casecmp('disable') == 0 }
+      end
+    end
+
+    def ensure_opposite?
+      properties.detect do |p|
+        p.is_ensure? && p.values.any? { |v| v.casecmp('enable') == 0 }
```
ferventcoder: I don't believe we need `ensure_opposite?` (the opposite functions) now that we are returning the `ensure_value` and `absent_value`.

ferventcoder:
This one is not going to pass specs given the changes would require the types to be regenerated. Will have to test locally.

build/dsc/resource.rb
```diff
@@ -82,7 +82,37 @@ def ensurable?
...
+    end
+
+    def ensure_value
+      properties.detect do |p|
+        p.is_ensure? && p.values.each do |v|
+          return 'present' if v.casecmp('present') == 0
+          return 'enable' if v.casecmp('enable') == 0
+        end
+      end
+    end
```
ferventcoder: I'm glad this code is only in the builder.

build/dsc/templates/dsc_type.rb.erb
```diff
@@ -5,6 +5,18 @@ Puppet::Type.newtype(:dsc_<%= resource.friendlyname.downcase %>) do
...
     defaultfor :operatingsystem => :windows
+
+  <% if resource.ensurable? -%>
+    def ensure_value
+      '<%= resource.ensure_value %>'
+    end
+   <% end -%>
+
+   <% if resource.absentable? -%>
+    def absent_value
```
ferventcoder: Some issue here causing the generation to be moved in too far. It comes out as:

~~~ruby
    def ensure_value
    'present'
  end
~~~

build/dsc/templates/dsc_type.rb.erb
```diff
@@ -5,6 +5,18 @@ Puppet::Type.newtype(:dsc_<%= resource.friendlyname.downcase %>) do
 
   provide :powershell, :parent => Puppet::Type.type(:base_dsc).provider(:powershell) do
     defaultfor :operatingsystem => :windows
+
+  <% if resource.ensurable? -%>
+    def ensure_value
```
ferventcoder: Some issue here causing the generation to be moved in too far. It comes out as:

~~~ruby
    def ensure_value
    'present'
  end
~~~

build/dsc/resource.rb
jpogran: Removed and updated dsc_type_spec.rb.erb

ferventcoder: These look like they are still here?

build/dsc/templates/dsc_type_spec.rb.erb
```diff
@@ -203,11 +203,11 @@ describe Puppet::Type.type(:dsc_<%= resource.friendlyname.downcase %>) do
 <%  end # unless ['msft_filedirectoryconfiguration'].... -%>
 <%  if resource.ensurable? -%>
 
-    describe "when dsc_ensure is 'present'" do
+    describe "when dsc_ensure is 'presemt'" do
```
ferventcoder: Looks like a misspelling was added inadvertently.

lib/puppet/provider/templates/invoke_dsc_resource.ps1.erb
```diff
@@ -21,7 +21,9 @@ $invokeParams = @{
 <% dsc_parameters.each do |p| -%>
     <%- name = p.name.to_s.gsub(/^dsc_/,'')
     value = format_dsc_value(p.value)
-    value = '\'present\'' if name == 'ensure' && dsc_invoke_method == 'test'
+    if name == 'ensure' && dsc_invoke_method == 'test'
+      value = "'#{resource.provider.ensure_value}'"
+    end
```
ferventcoder: Now that we've proven this out, let's flip it back like it was before with the if statement tacked on to setting the value.

Iristyle:
@jpogran can we add some notes to the commit message about the specifics?  i.e. which types use the different ensurable values?

I understand what's going on here, but the process seems a little dicey to me, because of a couple reasons:

* are we sure positive the semantics for `disable` and `absent` are the same?
* what happens if a resource is ever defined with both `disable` and `absent` for ensure values?
    * the logic right now is order dependent, rather than prioritizing one over the other

build/dsc/templates/dsc_type_spec.rb.erb
```diff
@@ -228,23 +228,22 @@ describe Puppet::Type.type(:dsc_<%= resource.friendlyname.downcase %>) do
 <%  end # if resource.ensurable? -%>
 <%  if resource.absentable? -%>
     describe "when dsc_ensure is 'absent'" do
-
       before(:each) do
-        dsc_<%= resource.friendlyname.downcase %>.original_parameters[:dsc_ensure] = 'absent'
-        dsc_<%= resource.friendlyname.downcase %>[:dsc_ensure] = 'absent'
+        dsc_<%= resource.friendlyname.downcase %>.original_parameters[:dsc_ensure] = '<%= resource.ensure_value %>'
```
Iristyle: This doesn't seem right to me, unless I've missed something. I think you mean `resource.absent_value` for this line and all the below...

ferventcoder:
> * are we sure positive the semantics for disable and absent are the same?

They are not. But we are not treating them the same in the way that we contact the DSC resource. We are just ensuring we pass through the proper thing to it, and with Puppet, we are ensuring that the items actually get called.

> * what happens if a resource is ever defined with both disable and absent for ensure values?
>   * the logic right now is order dependent, rather than prioritizing one over the other.

I think at that point we'll need to take a look at redesigning this.

build/dsc/templates/dsc_type_spec.rb.erb
```diff
@@ -228,23 +228,22 @@ describe Puppet::Type.type(:dsc_<%= resource.friendlyname.downcase %>) do
...
+        dsc_<%= resource.friendlyname.downcase %>[:dsc_ensure] = '<%= resource.ensure_value %>'
         @provider = described_class.provider(:powershell).new(dsc_<%= resource.friendlyname.downcase %>)
       end
 
       it "should update :ensure to :absent" do
-        expect(dsc_<%= resource.friendlyname.downcase %>[:ensure]).to eq(:absent)
+        expect(dsc_<%= resource.friendlyname.downcase %>[:ensure]).to eq(:present)
```
ferventcoder: This looks like another whoops.

build/dsc/templates/dsc_type_spec.rb.erb
ferventcoder: I believe you are correct. I did ask @jpogran to rebase #73 against this latest and ensure the specs pass. Which we can see they are not likely to. :)

build/dsc/templates/dsc_type.rb.erb
```diff
@@ -71,7 +84,7 @@ Puppet::Type.newtype(:dsc_<%= resource.friendlyname.downcase %>) do
 <%    end -%>
     validate do |value|
 <%    if property.name.downcase == 'ensure' -%>
-      resource[:ensure] = value.downcase
+      resource[:ensure] = provider.munge_ensure(value.downcase)
```
Iristyle: The `downcase` is not necessary here because `munge_ensure` is a case insensitive regex.  (It probably makes sense to handle any manipulation in `munge_ensure` as well)

build/dsc/resource.rb
```diff
@@ -82,7 +82,25 @@ def ensurable?
...
     def absentable?
       properties.detect do |p|
-        p.is_ensure? && p.values.any? { |v| v.casecmp('absent') == 0 }
+        p.is_ensure? && p.values.any? { |v| v.casecmp('absent') == 0 || v.casecmp('disable') == 0 }
+      end
+    end
+
+     def absent_value
+      properties.detect do |p|
+        p.is_ensure? && p.values.each do |v|
```
Iristyle: I think I would rewrite this part to future proof the schema parsing / prioritize `absent` (in the event `absent` and `disable` are both in the map)

```ruby
def absent_value
  properties.detect do |p|
    return 'absent' if p.is_ensure? && p.values.any { |v| v.casecmp('absent') == 0 }
    return 'disable' if p.is_ensure? && p.values.any { |v| v.casecmp('disable') == 0 }
  end
  throw 'Error processing MOF schema - could not determine valid ensure values'
end
```

Thoughts?

build/dsc/templates/dsc_type.rb.erb
jpogran: Agreed

build/dsc/resource.rb
jpogran: Agreed

lib/puppet/provider/base_dsc/powershell.rb
```diff
@@ -76,6 +76,13 @@ def munge_boolean(value)
     raise ArgumentError.new "invalid value: #{value}"
   end
 
+  def munge_ensure(value)
+    return 'absent' if value.downcase =~ (/^(absent|disable)$/i)
```
Iristyle: @jpogran shouldn't this look at the value that the type has defined for it's `absent_value`?

lib/puppet/provider/base_dsc/powershell.rb
```diff
@@ -76,6 +76,13 @@ def munge_boolean(value)
...
+    return 'present' if value.downcase =~ (/^(present|enable)$/i)
```
Iristyle: Likewise, shouldn't this look at the value that the type has defined for it's `present_value`?

build/dsc/resource.rb
ferventcoder: :+1:

lib/puppet/provider/base_dsc/powershell.rb
ferventcoder: Unfortunately we won't have access to the type since the type for `powershell.rb` is `base_dsc.rb`. However if we move this INTO the type in the section where it creates a provider named powershell that inherits from powershell.rb, we can do it there.

lib/puppet/provider/base_dsc/powershell.rb
ferventcoder: Same comment as above.

lib/puppet/provider/base_dsc/powershell.rb
ferventcoder: But then again, I would prefer for a munge_* to go into the type itself anyway. I think there was a reason we shied away from it and put this in `powershell.rb` but I don't remember it off the top of my head.

build/dsc/resource.rb
```diff
@@ -82,7 +82,26 @@ def ensurable?
 
     def absentable?
       properties.detect do |p|
-        p.is_ensure? && p.values.any? { |v| v.casecmp('absent') == 0 }
+        return 'absent' if p.is_ensure? && p.values.any? { |v| v.casecmp('absent') }
```
Iristyle: Just realized that the logic here is not quite right as well... this changes the previous `true` / `false` return values to something different.

Will fix in my updates.

ferventcoder: Not sure when the `return 'absent'` showed up.

ferventcoder: Likely a bad copy pasta over the weekend.

ferventcoder:
Closed in favor of #77.

PR #77, opened 2015-09-01
(MODULES-2479) Allow different ensure values / (FM-3578) Update DSC types

Iristyle:
Supercedes #73 and #76, which have been closed.

There were a handful of issues with the prior commits that have been fixed here:

* The `.git` files from submodules were not properly deleted as part of the new rake tasks - this resulted in issues trying to check in the vendored DSC resources
* Discovered a bug in how the `defaultto` value was being set for `ensurable` - addressed in b729ded
* Did quite a bit of rework to the changes necessary for `windowsoptionalfeature`, so that `enable` and `disable` are considered valid values.  Since Puppet already has the ability to use custom `ensure` values, leverage those and remove the custom code for `munge_ensure` which was not necessary.  Similarly, remove the additional methods for `ensure_value` and `absent_value` that were being put into the provider for each type, since they're not necessary.  Simplify test changes, so that tests don't break without re-running the codegen process.
* Separate out the updates that are WMF5 related from those that are part of the GitHub repo update
* Added in nearly 700 missing files, that I'm assuming were missed due to the `.git` file issue. - see https://github.com/Iristyle/puppetlabs-dsc/commit/3bea84b0139e1961d1a3968eda430550f22eb5bf for the specifics of those changes
* Deleted a number of files from the vendored DSC directory that should have been trashed given they're no longer part of the upstream PowerShell repos.

The validation for windowsoptionalfeature checks out when running locally on Windows 2008, but I didn't have the opportunity to fire up a client Windows version and actually test that it worked as expected.  If @jpogran or @ferventcoder could verify that prior to merge, that would be great.

There are a couple of additional things I didn't address here, that should be fixed in the future

* the rake clean tasks don't properly delete the vendored dsc resources (which result in extraneous files getting carried over)
* the rake tasks should exclude files that we don't need to vendor - `Samples/*`, `Examples/*`, `Tests/*`, `appveyor.yml`, etc


jpogran:
Tested working on win8.1x64

PR #88, opened 2015-09-03
(WIP - DO NOT MERGE)(MODULES-2178) PSCredential Type Generation

jpogran:


Iristyle:
Looks like this needs to be rebased @jpogran

Iristyle:
FYI @jpogran / @ferventcoder - I rebased and did a bunch of additional work to move things forward with this - it's in my branch at https://github.com/Iristyle/puppetlabs-dsc/commits/pr/88

What I discovered when using the embedded instance class name in https://github.com/Iristyle/puppetlabs-dsc/commit/538ac30709d0efaa888ee735a63652b67521554f (which I believe is the way we need to codegen around any complex DSC types) is that we have a few other resources that don't work as it stands:

* `MSFT_xAzureVMDscExtension.schema.mof` (of `dsc_xazurevmdscextension`) references `MSFT_KeyValuePair` with `String ConfigurationArgument` - this is a single hashtable entry with `key` and `value`
* `MSFT_xRemoteFile.schema.mof` (of `dsc_xremotefile`) references `MSFT_KeyValuePair` with `String Headers[]` - this corresponds to a an array of hashtable entries with `key` and `value`
* `MSFT_xVhdFileDirectory.schema.mof` (of `dsc_xvhdfile`) defines the type `MSFT_xFileDirectory`, so this one is complicated
* `MSFT_xWebsite.schema.mof` (of `dsc_xwebsite`) defines the type `MSFT_xWebBindingInformation`, so this one is complicated as well

I have initially addressed `MSFT_KeyValuePair` in https://github.com/Iristyle/puppetlabs-dsc/commit/8909b3eea8a4b774aff8309bec7a382d08a7c0cb and that was pretty straightforward, though I haven't properly addressed the Headers array mentioned above given it can contain multiple entries in an array though... so @jpogran that would be a good place to pickup where I've left off.

The other 2 guys are a bit more complicated and I think we may to punt on them to a new ticket?  Thoughts?


/cc @cowofevil - the 4 types mentioned in the bullets above are special cases for testing.

ferventcoder:
I like the changes for mof_type, that is much more deterministic.

Iristyle:
As a follow-up, I think I've gone slightly the wrong direction on `MSFT_KeyValuePair` and will be revising this accordingly.

Iristyle:
Superceded by #93

PR #89, opened 2015-09-03
(MODULES-2520) set readme parameters and clean up generated type docs

ferventcoder:
Clean up the documentation that can be queried with
puppet describe.

This should clean up https://forge.puppetlabs.com/puppetlabs/dsc/types into something more usable.

### DIFF
--- a/build/dsc.rake
+++ b/build/dsc.rake
@@ -26,6 +26,7 @@ namespace :dsc do
     Rake::Task['dsc:resources:import'].invoke unless File.exists?(default_dsc_resources_path)
     Rake::Task['dsc:types:clean'].invoke(dsc_module_path)
     Rake::Task['dsc:types:build'].invoke(dsc_module_path)
+    Rake::Task['dsc:types:document'].invoke(dsc_module_path)
   end
 
   desc "Cleanup all"
@@ -115,6 +116,15 @@ eod
       m.target_module_path = module_path
       msgs = m.build_dsc_types
       msgs.each{|m| puts "#{m}"}
+      dsc_types = m.get_dsc_types
+    end
+
+    desc "Document #{item_name}"
+    task :document, [:module_path] do |t, args|
+      module_path = args[:module_path] || default_dsc_module_path
+      m = Dsc::Manager.new
+      m.target_module_path = module_path
+      m.document_types("#{default_dsc_module_path}/Types.md",m.get_dsc_types)
     end
 
     desc "Cleanup #{item_name}"
--- a/build/dsc/manager.rb
+++ b/build/dsc/manager.rb
@@ -148,6 +148,25 @@ module Dsc
       type_pathes
     end
 
+    def get_dsc_types
+      dsc_types = []
+      resources.each do |resource|
+        dsc_types << "dsc_#{resource.friendlyname.downcase}"
+      end
+      dsc_types
+    end
+
+    def document_types(markdown_file_path,dsc_types)
+      puts "Creating #{markdown_file_path}"
+      File.open("#{markdown_file_path}", 'w+') do |file|
+        file.write("## Resource Types included with dsc module\n")
+        file.write("For any system you have this installed on you can use\n`Puppet describe typename` for more information.\n\n")
+        dsc_types.each do |dsc_type|
+          file.write("* #{dsc_type}\n")
+        end
+      end
+    end
+
     def clean_dsc_types
       puppet_type_path = "#{@target_module_path}/#{@puppet_type_subpath}"
       clean_folder(["#{puppet_type_path}/dsc_*.rb"])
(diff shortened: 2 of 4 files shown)


ferventcoder:
This has been tested building types, but I didn't want to include that here until we are ready for that (or let the types be built later).

build/dsc/templates/dsc_type.rb.erb
```diff
@@ -63,9 +87,9 @@ Puppet::Type.newtype(:dsc_<%= resource.friendlyname.downcase %>) do
   # IsMandatory:  <%= property.required? ? 'True' : 'False' %>
   # Values:       <%= property.values ? property.values : 'None' %>
   newparam(:dsc_<%= property.name.downcase %><% if property.array? %>, :array_matching => :all<% end %>) do
-<%    if property.description -%>
-    desc "<%= property.description %>"
-<%    end -%>
+    desc "<%= property.name %><% if property.description -%> - <%= property.description %><% end %>
+      Expects input to be a <%= property.type %>.
```
Iristyle: `property.type` is more or less the MOF type, so it might not be what we want to document for Puppet users

I think if we want to document the Puppet type here, we will need to build some sort of conversion table based on https://github.com/puppetlabs/puppetlabs-dsc/blob/master/build/dsc/property.rb#L75-L105

ferventcoder: Yep, let's leave that bit out for now.

ferventcoder:
@Iristyle @jpogran was there anything additional to do here?

ferventcoder:
This should be validated that the move to property instead of parameter doesn't break anything.

Iristyle:
I think this looks good, but will do a once over in more detail prior to merge.

ferventcoder:
Needs a rebase.

build/dsc/templates/dsc_type.rb.erb
```diff
@@ -55,8 +55,17 @@ Puppet::Type.newtype(:dsc_<%= resource.friendlyname.downcase %>) do
 
   @doc = %q{
     The DSC <%= resource.friendlyname%> resource type.
-    Originally generated from the following schema.mof file:
-      <%= resource.relative_mof_path %>
+    Automatically generated from
+    '<%= resource.relative_mof_path.gsub('import/dsc_resources/','') %>'
```
jpogran: I like the path used here, immediately apparent which module it comes from without the implementation details of where it was vendored in the source

build/dsc/templates/dsc_type.rb.erb
```diff
@@ -83,12 +110,31 @@ Puppet::Type.newtype(:dsc_<%= resource.friendlyname.downcase %>) do
...
+      fail 'dscmeta_module_name is read-only'
+    end
   end
 <%  if resource.ps_module.name != 'PSDesiredStateConfiguration' %>
-  newparam(:dscmeta_module_version) do
-    defaultto "<%= resource.ps_module.version %>"
+  newproperty(:dscmeta_module_version) do
+    desc "A read-only value for the DSC Module Version ('<%= resource.ps_module.version %>').
+      This is the supported version of the PowerShell module that this
+      type was built on. When puppet runs this resource, it will explicitly
```
jpogran: Capital Puppet?

build/dsc/templates/dsc_type.rb.erb
```diff
@@ -130,9 +176,8 @@ Puppet::Type.newtype(:dsc_<%= resource.friendlyname.downcase %>) do
       <%= type_map.to_s %>
     end
 <%    end -%>
-<%    if property.description -%>
-    desc "<%= property.description %>"
-<%    end -%>
+    desc "<%= property.name %><% if property.description -%> - <%= property.description %><% end %>
+      <% if property.values -%>Valid values are <%= property.values.join(', ') %><% end -%>"
```
jpogran: The only nit here is if there isn't a range of values, there is a newline plus empty space here. Example:

```
    desc "Name
      "
```

ferventcoder: That actually seems easy to fixup. :)

build/dsc/templates/dsc_type.rb.erb
ferventcoder: yes.

build/dsc/manager.rb
```diff
@@ -148,6 +148,25 @@ def build_dsc_types
...
+    def document_types(markdown_file_path,dsc_types)
+      puts "Creating #{markdown_file_path}"
+      File.open("#{markdown_file_path}", 'w+') do |file|
+        file.write("## Resource Types included with dsc module\n")
+        file.write("For any system you have this installed on you can use\n`Puppet describe typename` for more information.\n\n")
+        dsc_types.each do |dsc_type|
+          file.write("* #{dsc_type}\n")
+        end
+      end
+    end
```
ferventcoder: generates the type.md

build/dsc.rake
```diff
@@ -115,6 +116,15 @@ eod
       m.target_module_path = module_path
       msgs = m.build_dsc_types
       msgs.each{|m| puts "#{m}"}
+      dsc_types = m.get_dsc_types
```
ferventcoder: @Iristyle oops. Just noticed this one. I can clean it up, it's removing a single line.

PR #99, opened 2015-09-16
(MODULES-2521) Confine to PowerShell version 5

jpogran:
Confine the module to PowerShell version 5 as the functionality we require only exists in that version

### DIFF
--- a/build/dsc/templates/dsc_type.rb.erb
+++ b/build/dsc/templates/dsc_type.rb.erb
@@ -5,6 +5,7 @@ Puppet::Type.newtype(:dsc_<%= resource.friendlyname.downcase %>) do
   require Pathname.new(__FILE__).dirname + '../../puppet_x/puppetlabs/dsc_type_helpers'
 
   provide :powershell, :parent => Puppet::Type.type(:base_dsc).provider(:powershell) do
+    confine :true => (Gem::Version.new(Facter.value(:powershell_version)) >= Gem::Version.new('5.0.10240.16384'))
     defaultfor :operatingsystem => :windows
   end
 <%  ignored = ['MSFT_KeyValuePair', 'MSFT_Credential']
--- a/lib/facter/powershell_version.rb
+++ b/lib/facter/powershell_version.rb
@@ -1,0 +1,11 @@
+require 'pathname'
+require Pathname.new(__FILE__).dirname + '../' + 'puppet_x/puppetlabs/powershell_version'
+
+Facter.add(:powershell_version) do
+  setcode do
+    if Puppet::Util::Platform.windows?
+      version = PuppetX::PowerShell::PowerShellVersion.version
+      version
+    end
+  end
+end
--- a/lib/puppet/provider/base_dsc/powershell.rb
+++ b/lib/puppet/provider/base_dsc/powershell.rb
@@ -36,6 +36,8 @@ EOT
   end
 
   def exists?
+    version = Facter.value(:powershell_version)
+    Puppet.debug "PowerShell Version: #{version}"
     script_content = ps_script_content('test')
     Puppet.debug "\n" + script_content
     output = powershell(powershell_args, script_content)
--- a/lib/puppet/type/dsc_archive.rb
+++ b/lib/puppet/type/dsc_archive.rb
@@ -5,6 +5,7 @@ Puppet::Type.newtype(:dsc_archive) do
   require Pathname.new(__FILE__).dirname + '../../puppet_x/puppetlabs/dsc_type_helpers'
 
   provide :powershell, :parent => Puppet::Type.type(:base_dsc).provider(:powershell) do
+    confine :true => (Gem::Version.new(Facter.value(:powershell_version)) >= Gem::Version.new('5.0.10240.16384'))
     defaultfor :operatingsystem => :windows
   end
 
--- a/lib/puppet/type/dsc_environment.rb
+++ b/lib/puppet/type/dsc_environment.rb
@@ -5,6 +5,7 @@ Puppet::Type.newtype(:dsc_environment) do
   require Pathname.new(__FILE__).dirname + '../../puppet_x/puppetlabs/dsc_type_helpers'
 
   provide :powershell, :parent => Puppet::Type.type(:base_dsc).provider(:powershell) do
+    confine :true => (Gem::Version.new(Facter.value(:powershell_version)) >= Gem::Version.new('5.0.10240.16384'))
     defaultfor :operatingsystem => :windows
   end
 
--- a/lib/puppet/type/dsc_file.rb
+++ b/lib/puppet/type/dsc_file.rb
@@ -5,6 +5,7 @@ Puppet::Type.newtype(:dsc_file) do
   require Pathname.new(__FILE__).dirname + '../../puppet_x/puppetlabs/dsc_type_helpers'
 
   provide :powershell, :parent => Puppet::Type.type(:base_dsc).provider(:powershell) do
+    confine :true => (Gem::Version.new(Facter.value(:powershell_version)) >= Gem::Version.new('5.0.10240.16384'))
     defaultfor :operatingsystem => :windows
   end
 
--- a/lib/puppet/type/dsc_group.rb
+++ b/lib/puppet/type/dsc_group.rb
@@ -5,6 +5,7 @@ Puppet::Type.newtype(:dsc_group) do
   require Pathname.new(__FILE__).dirname + '../../puppet_x/puppetlabs/dsc_type_helpers'
 
   provide :powershell, :parent => Puppet::Type.type(:base_dsc).provider(:powershell) do
+    confine :true => (Gem::Version.new(Facter.value(:powershell_version)) >= Gem::Version.new('5.0.10240.16384'))
     defaultfor :operatingsystem => :windows
   end
 
--- a/lib/puppet/type/dsc_log.rb
+++ b/lib/puppet/type/dsc_log.rb
@@ -5,6 +5,7 @@ Puppet::Type.newtype(:dsc_log) do
   require Pathname.new(__FILE__).dirname + '../../puppet_x/puppetlabs/dsc_type_helpers'
 
   provide :powershell, :parent => Puppet::Type.type(:base_dsc).provider(:powershell) do
+    confine :true => (Gem::Version.new(Facter.value(:powershell_version)) >= Gem::Version.new('5.0.10240.16384'))
     defaultfor :operatingsystem => :windows
   end
 
--- a/lib/puppet/type/dsc_package.rb
+++ b/lib/puppet/type/dsc_package.rb
@@ -5,6 +5,7 @@ Puppet::Type.newtype(:dsc_package) do
   require Pathname.new(__FILE__).dirname + '../../puppet_x/puppetlabs/dsc_type_helpers'
 
   provide :powershell, :parent => Puppet::Type.type(:base_dsc).provider(:powershell) do
+    confine :true => (Gem::Version.new(Facter.value(:powershell_version)) >= Gem::Version.new('5.0.10240.16384'))
     defaultfor :operatingsystem => :windows
   end
 
--- a/lib/puppet/type/dsc_registry.rb
+++ b/lib/puppet/type/dsc_registry.rb
@@ -5,6 +5,7 @@ Puppet::Type.newtype(:dsc_registry) do
   require Pathname.new(__FILE__).dirname + '../../puppet_x/puppetlabs/dsc_type_helpers'
 
   provide :powershell, :parent => Puppet::Type.type(:base_dsc).provider(:powershell) do
+    confine :true => (Gem::Version.new(Facter.value(:powershell_version)) >= Gem::Version.new('5.0.10240.16384'))
     defaultfor :operatingsystem => :windows
   end
 
--- a/lib/puppet/type/dsc_script.rb
+++ b/lib/puppet/type/dsc_script.rb
@@ -5,6 +5,7 @@ Puppet::Type.newtype(:dsc_script) do
   require Pathname.new(__FILE__).dirname + '../../puppet_x/puppetlabs/dsc_type_helpers'
 
   provide :powershell, :parent => Puppet::Type.type(:base_dsc).provider(:powershell) do
+    confine :true => (Gem::Version.new(Facter.value(:powershell_version)) >= Gem::Version.new('5.0.10240.16384'))
     defaultfor :operatingsystem => :windows
   end
 
--- a/lib/puppet/type/dsc_service.rb
+++ b/lib/puppet/type/dsc_service.rb
@@ -5,6 +5,7 @@ Puppet::Type.newtype(:dsc_service) do
   require Pathname.new(__FILE__).dirname + '../../puppet_x/puppetlabs/dsc_type_helpers'
 
   provide :powershell, :parent => Puppet::Type.type(:base_dsc).provider(:powershell) do
+    confine :true => (Gem::Version.new(Facter.value(:powershell_version)) >= Gem::Version.new('5.0.10240.16384'))
     defaultfor :operatingsystem => :windows
   end
 
--- a/lib/puppet/type/dsc_user.rb
+++ b/lib/puppet/type/dsc_user.rb
@@ -5,6 +5,7 @@ Puppet::Type.newtype(:dsc_user) do
   require Pathname.new(__FILE__).dirname + '../../puppet_x/puppetlabs/dsc_type_helpers'
 
   provide :powershell, :parent => Puppet::Type.type(:base_dsc).provider(:powershell) do
+    confine :true => (Gem::Version.new(Facter.value(:powershell_version)) >= Gem::Version.new('5.0.10240.16384'))
     defaultfor :operatingsystem => :windows
   end
 
--- a/lib/puppet/type/dsc_waitforall.rb
+++ b/lib/puppet/type/dsc_waitforall.rb
@@ -5,6 +5,7 @@ Puppet::Type.newtype(:dsc_waitforall) do
   require Pathname.new(__FILE__).dirname + '../../puppet_x/puppetlabs/dsc_type_helpers'
 
   provide :powershell, :parent => Puppet::Type.type(:base_dsc).provider(:powershell) do
+    confine :true => (Gem::Version.new(Facter.value(:powershell_version)) >= Gem::Version.new('5.0.10240.16384'))
     defaultfor :operatingsystem => :windows
   end
 
--- a/lib/puppet/type/dsc_waitforany.rb
+++ b/lib/puppet/type/dsc_waitforany.rb
@@ -5,6 +5,7 @@ Puppet::Type.newtype(:dsc_waitforany) do
   require Pathname.new(__FILE__).dirname + '../../puppet_x/puppetlabs/dsc_type_helpers'
 
   provide :powershell, :parent => Puppet::Type.type(:base_dsc).provider(:powershell) do
+    confine :true => (Gem::Version.new(Facter.value(:powershell_version)) >= Gem::Version.new('5.0.10240.16384'))
     defaultfor :operatingsystem => :windows
   end
 
--- a/lib/puppet/type/dsc_waitforsome.rb
+++ b/lib/puppet/type/dsc_waitforsome.rb
@@ -5,6 +5,7 @@ Puppet::Type.newtype(:dsc_waitforsome) do
   require Pathname.new(__FILE__).dirname + '../../puppet_x/puppetlabs/dsc_type_helpers'
 
   provide :powershell, :parent => Puppet::Type.type(:base_dsc).provider(:powershell) do
+    confine :true => (Gem::Version.new(Facter.value(:powershell_version)) >= Gem::Version.new('5.0.10240.16384'))
     defaultfor :operatingsystem => :windows
   end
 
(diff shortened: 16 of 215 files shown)


lib/facter/powershell_version.rb
```diff
@@ -0,0 +1,12 @@
+require 'win32/registry'
```
ferventcoder: The require should move after the confine.

ferventcoder: In short, this doesn't exist in non-Windows rubies and will cause issues being here.

lib/puppet/provider/base_dsc/powershell.rb
```diff
@@ -3,8 +3,8 @@
 require 'json'
 
 Puppet::Type.type(:base_dsc).provide(:powershell) do
-  confine :operatingsystem => :windows
-  defaultfor :operatingsystem => :windows
+  confine      :operatingsystem => :windows, :true => (Puppet[:powershell_version] == '5')
+  defaultfor   :operatingsystem => :windows
```
ferventcoder: These probably should go into the actual type provider overrides.

lib/facter/powershell_version.rb
```diff
@@ -0,0 +1,12 @@
...
+
+Facter.add(:powershell_version) do
+  confine :operatingsystem => :windows
+  setcode do
+    # HKLM:\SOFTWARE\Microsoft\PowerShell\3\PowerShellEngine
+    # Win32::Registry::HKEY_LOCAL_MACHINE.open('SOFTWARE\Microsoft\PowerShell\3\PowerShellEngine') do |reg|
+    #   reg_val = reg['PowerShellVersion']
+    # end
```
cowofevil: Dead code should be removed.

lib/facter/powershell_version.rb
jpogran: Fixed, thanks

lib/facter/powershell_version.rb
jpogran: If it's dead yes, but this was commented out to figure out why confine wasn't working. We are actually going to have to figure out the version on the client, so we need this.

lib/puppet/provider/base_dsc/powershell.rb
jpogran: Correct

jpogran:
This still needs one more commit to build all the types with the new confine.

lib/facter/powershell_version.rb
```diff
@@ -0,0 +1,49 @@
...
+    if powershell_version = Facter.value(:powershell_version)
+      powershell_version.to_s.split('.')[1]
+    end
+  end
+end
+
+Facter.add(:powershell_patch_version) do
+  confine :operatingsystem => :windows
+  setcode do
+    if pe_version = Facter.value(:powershell_version)
```
ferventcoder: why `pe_version`?

lib/facter/powershell_version.rb
```diff
@@ -0,0 +1,49 @@
+Facter.add(:powershell_installed) do
+ confine :operatingsystem => :windows
+ require 'win32/registry'
```
ferventcoder: This may require checking to see if we are in Windows. `if Puppet::Util::Platform.windows?`  - https://github.com/chocolatey/puppet-chocolatey/blob/master/lib/facter/choco_install_path.rb#L2-L19

ferventcoder:
Specs - https://github.com/chocolatey/puppet-chocolatey/blob/master/spec/unit/facter/choco_install_path_spec.rb

lib/facter/powershell_version.rb
```diff
@@ -0,0 +1,49 @@
...
+ setcode do
+  begin
+    Win32::Registry::HKEY_LOCAL_MACHINE.open('SOFTWARE\Microsoft\PowerShell\1\PowerShellEngine', ::Win32::Registry::KEY_READ)
```
ferventcoder: This key will be in both x64 and Wow6432Node, so I don't know if we need to turn off registry redirection. Prolly for safety we should.

ferventcoder: The flags `Win32::Registry::KEY_READ | 0x100`

lib/facter/powershell_version.rb
```diff
@@ -0,0 +1,49 @@
...
+    false
+  end
+ end
+end
+
+Facter.add(:powershell_version) do
+  confine :operatingsystem => :windows, :powershell_installed => true
+  require 'win32/registry'
+  setcode do
+    Win32::Registry::HKEY_LOCAL_MACHINE.open('SOFTWARE\Microsoft\PowerShell\3\PowerShellEngine') do |reg|
```
ferventcoder: This key will be in both x64 and Wow6432Node, so I don't know if we need to turn off registry redirection (the flags `Win32::Registry::KEY_READ | 0x100`). Prolly for safety we should.

ferventcoder: This check could actually run against Posh v2. I'd surround this with begin/rescue.

ferventcoder: Example https://github.com/chocolatey/puppet-chocolatey/blob/master/lib/facter/choco_install_path.rb#L8-L15

ferventcoder: This check may need to have a fallback to `PowerShell\1\PowerShellEngine` when the first call fails. Then maybe another fallback?

lib/facter/powershell_version.rb
```diff
@@ -0,0 +1,49 @@
...
+end
+
+Facter.add(:powershell_patch_version) do
+  confine :operatingsystem => :windows
+  setcode do
+    if pe_version = Facter.value(:powershell_version)
+      pe_version.to_s.split('.')[2]
+    end
+  end
+end
```
ferventcoder: I imagine you are going to split this to multiple files.

build/dsc/templates/dsc_type.rb.erb
```diff
@@ -4,6 +4,7 @@ Puppet::Type.newtype(:dsc_<%= resource.friendlyname.downcase %>) do
   require Pathname.new(__FILE__).dirname + '../../' + 'puppet/type/base_dsc'
 
   provide :powershell, :parent => Puppet::Type.type(:base_dsc).provider(:powershell) do
+    confine :powershell_major_version => '5'
```
Iristyle: Should we also do a minor / patch version here?  We know that we want at least the production preview for the LCM changes.

Iristyle:
Note that this still doesn't work until some underlying changes have been made to move helper methods off of the provider.  Otherwise the confine will correctly set the `provider` to `nil` and try to access `provider.XXX` and then throw an exception.

These changes are going to make it into a quick separate PR.

Iristyle:
Requires #102 to be merged first

build/dsc/templates/dsc_type.rb.erb
jpogran: Had gone back and forth on this. If we tie to minor and patch, we will have to release a new version when they do, even though we might not have to have any changes ourselves. The benefit of ensuring user is always on proper version outweighs this, so implementing this

ferventcoder:
The facter powershell_version.rb still needs split into several files, one per fact.

lib/facter/powershell_version.rb
```diff
@@ -0,0 +1,45 @@
...
+      require 'win32/registry'
+      version = nil
+      begin
+        access_type = Win32::Registry::KEY_READ | 0x100
+        hive        = Win32::Registry::HKEY_LOCAL_MACHINE
+        hive.open('SOFTWARE\Microsoft\PowerShell\3\PowerShellEngine', access_type) do |reg|
+          version = reg['PowerShellVersion']
+        end
+      rescue
+        version = nil
```
ferventcoder: As [mentioned before](https://github.com/puppetlabs/puppetlabs-dsc/pull/99#discussion_r40114588), this check may need to have a fallback to PowerShell\1\PowerShellEngine when the first call fails. Then maybe another fallback in case PowerShell is not installed at all. Although I'm not really sure how often we'd hit the third scenario given we are moving into places where POSH is already installed by default.

ferventcoder: This is because this fact is before a confine to a posh version, so it could be run against PowerShell v2, which will be in the `1` key and not the `3` key.

build/dsc/templates/dsc_type.rb.erb
ferventcoder: I thought we were still going to allow folks to use this module with the optional `lcm_config`? Confining further seems like an anti-pattern to me.

ferventcoder:
I'm not sure I agree with further confinement than the major version.

build/dsc/templates/dsc_type.rb.erb
Iristyle: Users are still allowed to use `lcm_config` if they want, sure.

The sentiment here though is that we're trying to set a minimum allowable version (not picking an explicit version). My main concern is ensuring end users are at least on the supported production preview release of WMF5, which is what we test against... and not some prior release of WMF5.

I think it's also nice to be inline with our documentation, which states `lcm_config` is optional.  On earlier WMF5 releases, `lcm_config` is mandatory.

build/dsc/templates/dsc_type.rb.erb
```diff
@@ -4,6 +4,7 @@ Puppet::Type.newtype(:dsc_<%= resource.friendlyname.downcase %>) do
   require Pathname.new(__FILE__).dirname + '../../' + 'puppet/type/base_dsc'
 
   provide :powershell, :parent => Puppet::Type.type(:base_dsc).provider(:powershell) do
+    confine :operatingsystem => :windows, :powershell_version => '5.0.10514.6'
```
jpogran: This breaks all tests.

Iristyle: I'm not quite sure how that would work without custom comparisons, given
that's not a version object.

Something like http://stackoverflow.com/a/3064161 should work though

jpogran: The comparision works. I think the tests aren't creating a provider and thats why they are failing.

Iristyle: You can stub out the return value of that fact on non Windows and tests
should pass again.

I haven't looked at your latest code, since I'm on phone - but if the
comparison is just a string, it might be "working" in some cases, but you
can't compare versions like that, in a general sense.

Iristyle: Here's an example of what I mean:

```
[1] pry(main)> '5.0.10514.6' > '5.0.2.0'
=> false

[2] pry(main)> Gem::Version.new('5.0.10514.6') > Gem::Version.new('5.0.2.0')
=> true
```

lib/facter/powershell_version.rb
```diff
@@ -0,0 +1,76 @@
+Facter.add(:powershell_version) do
+  setcode do
+    if Puppet::Util::Platform.windows?
+      version = Facter.value(:powershell_three_version) || Facter.value(:powershell_one_version)
```
Iristyle: I think we're OK with wrapping this in a `Gem::Version.new()`

lib/facter/powershell_version.rb
```diff
@@ -0,0 +1,76 @@
...
+        end
+      rescue
+        version = nil
+      end
+    end
+    version
+  end
+end
+
+Facter.add(:powershell_major_version) do
```
Iristyle: If we return a `Gem::Version` as `:powershell_version` then I don't think these bits are necessary.

lib/facter/powershell_version.rb
Iristyle: But I would also double check to make sure that Facter doesn't turn it into a `string` by the time we use it in our `:confine`.

build/dsc/templates/dsc_type.rb.erb
Iristyle: As long as we can return a `Gem::Version` here, I think our confine can be rewrtiten as:

```ruby
confine :operatingsystem => :windows, :powershell_version => Gem::Version.new('5.0.10514.6')
```

But I've only returned basic types like strings and numerics from facts previously, so double check that Facter will preserve the `Gem::Version` type, and doesn't stringify it.

If it is stringifying it, we can rewrite the confine another way to turn the fact into a `Gem::Version` for the sake of comparison.

ferventcoder:
#102 was merged into master at https://github.com/puppetlabs/puppetlabs-dsc/commit/312c4d72c8787bb50bec4309eeca00f2a05f62dd

build/dsc/templates/dsc_type.rb.erb
ferventcoder: So basically what you are stating is that no user of Windows 10 is going to be able to use this.

ferventcoder: I'm just clarifying, because the versions are different and it is not the Production Preview release. Further, the users of Windows 10 are not allowed to upgrade to the Production Preview.

ferventcoder: Forgot the `;)` - added just now.

You may need to work on this fact then, this is not based in minimums, but exact matches at the moment.

build/dsc/templates/dsc_type.rb.erb
ferventcoder: If this is intended for minimum allowable version, I think we may need a better way of handling the confine.

build/dsc/templates/dsc_type.rb.erb
ferventcoder: > I think it's also nice to be inline with our documentation, which states lcm_config is optional. On earlier WMF5 releases, lcm_config is mandatory.

@jpogran and I talked about whether to add the line about it being mandatory still for Windows 10. He talked me out of adding it, but it's worth noting that it is still mandatory for Windows 10 users until RTM comes out.

build/dsc/templates/dsc_type.rb.erb
ferventcoder: FWIW, this is the minimum version that should be supported:
image

build/dsc/templates/dsc_type.rb.erb
ferventcoder: Windows 10 PowerShell version is `5.0.10240.16384`.

build/dsc/templates/dsc_type.rb.erb
Iristyle: Yeah, Windows 10.. sigh.  Thanks for the reminder.  Did you happen to check LCM behavior with `Invoke-DscResource` on that version?

So we have `5.0.10514.6`, unless on Windows 10... in which case we have `5.0.10240.16384`.  And no Windows Updates have shown up yet for it either?

I wonder if our best bet is to:

* Confine to `5.0.10240.16384` or higher (for now)
* Emit a warning from Puppet about the LCM business if < `5.0.10514.6` (for now)
* Add a small note to README (for now)

Then hopefully we can change the confine to `5.0.10514.6` pretty soon, and update the README by the time the module hits 1.0 / supported.

Thoughts @jpogran / @ferventcoder ?

Iristyle: Also, I should have elaborated on the confine comparison @ferventcoder.  Thanks for pointing that out.  We do need to rewrite the confine a bit since the current fact comparison is a match.  Assuming the fact can return a `Gem::Version`, then this:

```ruby
confine :true => (Facter[:powershell_version] >= Gem::Version.new('5.0.10240.16384'))
```

If its a string, then:

```ruby
confine :true => (Gem::Version.new(Facter[:powershell_version]) >= Gem::Version.new('5.0.10240.16384'))
```

Hope thats clearer.

ferventcoder: :+1:

jpogran: I don't know what the default `RefreshMode` is in the Windows 10 WMF release, as the VM i have isn't fresh, but it does prevent using `Invoke-DscResource` when `RefreshMode` is not `Disabled`

~~~
[jpogran@foo-test]$ Invoke-DscResource -Name file -Method set -Property @{ ensure = 'present'; destinationpath = 'c:\foo.txt'; contents= 'bar' } -Verbose
VERBOSE: Performing the operation "Invoke-CimMethod: Resourceset" on target "MSFT_DSCLocalConfigurationManager".
VERBOSE: Perform operation 'Invoke CimMethod' with following parameters, ''methodName' = Resourceset,'className' =
MSFT_DSCLocalConfigurationManager,'namespaceName' = root/Microsoft/Windows/DesiredStateConfiguration'.
VERBOSE: An LCM method call arrived from computer FOO with user sid S-1-5-21-1787757173-3380529767-833788528-1001.
VERBOSE: Operation 'Invoke CimMethod' complete.
Invoke-DscResource : Invoke-DscResource cmdlet is not supported in Push mode. Please set the LCM RefreshMode to DISABLED.
At line:1 char:1
+ Invoke-DscResource -Name file -Method set -Property @{ ensure = 'pres ...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotImplemented: (root/Microsoft/...gurationManager:String) [Invoke-DscResource], CimException
    + FullyQualifiedErrorId : MI RESULT 7,Microsoft.PowerShell.DesiredStateConfiguration.Commands.InvokeDscResourceMethodCommand
~~

lib/facter/powershell_version.rb
jpogran: Facter does not appear to turn it into a `string`, but keep it as a `Gem::Version`. However it does not accept that object as a valid value:

~~~
Warning: Facter: Fact resolution fact='powershell_version', resolution='<anonymous>' resolved to an invalid value: Expected 5.0.10514.6 to be one of [
Integer, Float, TrueClass, FalseClass, NilClass, String, Array, Hash], but was Gem::Version
~~~

build/dsc/templates/dsc_type.rb.erb
ferventcoder: Figures

Iristyle:
Windows spec is failing at

```
Failures:
 
  1) PuppetX::PowerShell::PowerShellVersion detecting versions should detect powershell 5 if installed 
     Failure/Error: subject.expects(:version).returns('5') 
     Mocha::ExpectationError: 
       not all expectations were satisfied 
       unsatisfied expectations: 
       - expected exactly once, not yet invoked: #<PuppetX::PowerShell::PowerShellVersion:0xbb39848>.version(any_parameters) 
       satisfied expectations: 
       - allowed any number of times, not yet invoked: #<Puppet::Util::Feature:0x324f240>.root?(any_parameters) 
     # ./spec/unit/puppet_x/puppetlabs/powershell_version_spec.rb:9:in `block (3 levels) in <top (required)>' 
     # C:/Ruby193/lib/ruby/gems/1.9.1/gems/rspec-core-2.14.8/lib/rspec/core/example.rb:114:in `instance_eval' 
     # C:/Ruby193/lib/ruby/gems/1.9.1/gems/rspec-core-2.14.8/lib/rspec/core/example.rb:114:in `block in run' 
     # C:/Ruby193/lib/ruby/gems/1.9.1/gems/rspec-core-2.14.8/lib/rspec/core/example.rb:254:in `with_around_each_hooks' 
     # C:/Ruby193/lib/ruby/gems/1.9.1/gems/rspec-core-2.14.8/lib/rspec/core/example.rb:111:in `run' 
     # C:/Ruby193/lib/ruby/gems/1.9.1/gems/rspec-core-2.14.8/lib/rspec/core/example_group.rb:390:in `block in run_examples' 
     # C:/Ruby193/lib/ruby/gems/1.9.1/gems/rspec-core-2.14.8/lib/rspec/core/example_group.rb:386:in `map' 
     # C:/Ruby193/lib/ruby/gems/1.9.1/gems/rspec-core-2.14.8/lib/rspec/core/example_group.rb:386:in `run_examples' 
     # C:/Ruby193/lib/ruby/gems/1.9.1/gems/rspec-core-2.14.8/lib/rspec/core/example_group.rb:371:in `run' 
     # C:/Ruby193/lib/ruby/gems/1.9.1/gems/rspec-core-2.14.8/lib/rspec/core/example_group.rb:372:in `block in run' 
     # C:/Ruby193/lib/ruby/gems/1.9.1/gems/rspec-core-2.14.8/lib/rspec/core/example_group.rb:372:in `map' 
     # C:/Ruby193/lib/ruby/gems/1.9.1/gems/rspec-core-2.14.8/lib/rspec/core/example_group.rb:372:in `run' 
     # C:/Ruby193/lib/ruby/gems/1.9.1/gems/rspec-core-2.14.8/lib/rspec/core/command_line.rb:28:in `block (2 levels) in run' 
     # C:/Ruby193/lib/ruby/gems/1.9.1/gems/rspec-core-2.14.8/lib/rspec/core/command_line.rb:28:in `map' 
     # C:/Ruby193/lib/ruby/gems/1.9.1/gems/rspec-core-2.14.8/lib/rspec/core/command_line.rb:28:in `block in run' 
     # C:/Ruby193/lib/ruby/gems/1.9.1/gems/rspec-core-2.14.8/lib/rspec/core/reporter.rb:58:in `report' 
     # C:/Ruby193/lib/ruby/gems/1.9.1/gems/rspec-core-2.14.8/lib/rspec/core/command_line.rb:25:in `run' 
     # C:/Ruby193/lib/ruby/gems/1.9.1/gems/rspec-core-2.14.8/lib/rspec/core/runner.rb:80:in `run' 
     # C:/Ruby193/lib/ruby/gems/1.9.1/gems/rspec-core-2.14.8/lib/rspec/core/runner.rb:17:in `block in autorun' 
```

ferventcoder:
I would still love a spec on the confines to determine it is working correctly.

jakeballard:
Just a note, incase anyone else encounters it, this seems to be failing in Vagrant currently due this, [FACT-106](https://tickets.puppetlabs.com/browse/FACT-1276), bug in facter. The bug is preventing facter from running the custom fact for the powershell_version, so the version check fails. 

`==> default: Debug: Facter: custom fact "powershell_version" was not found.`
...
`==> default: Error: Could not find a suitable provider for dsc_windowsfeature`

Perhaps it is more clear when the fact actually runs, but it was not clear what it meant by not finding 'a suitable provider'. It would help if it also stated what it was looking for.

ferventcoder:
@jakeballard thanks for the heads up.

PR #102, opened 2015-09-22
(maint) PowerShell provider helpers -> puppet_x

Iristyle:
- Previously the munge_boolean helper was a part of the PowerShell
   provider implementation, even though it had nothing to do with
   provider specifics and is effectively plain old Ruby code.

   Refactor that method to a helper class, that will be expanded in
   future ticket MODULES-2178, and is also necessary for MODULES-2521
   which confines the provider to only work against PowerShell 5 or
   higher.

   For the confine to work properly, either calls to the provider must
   be guarded in the event of a nil provider instance, or those methods
   must be moved to an area that is accessible regardless of provider.

   In this instance, it makes more sense to introduce Ruby helpers not
   bound to the provider.

### DIFF
--- a/build/dsc/templates/dsc_type.rb.erb
+++ b/build/dsc/templates/dsc_type.rb.erb
@@ -2,6 +2,7 @@ require 'pathname'
 
 Puppet::Type.newtype(:dsc_<%= resource.friendlyname.downcase %>) do
   require Pathname.new(__FILE__).dirname + '../../' + 'puppet/type/base_dsc'
+  require Pathname.new(__FILE__).dirname + '../../puppet_x/puppetlabs/dsc_type_helpers'
 
   provide :powershell, :parent => Puppet::Type.type(:base_dsc).provider(:powershell) do
     defaultfor :operatingsystem => :windows
@@ -31,7 +32,7 @@ Puppet::Type.newtype(:dsc_<%= resource.friendlyname.downcase %>) do
     newvalues(true, false)
 
     munge do |value|
-      provider.munge_boolean(value.to_s)
+      PuppetX::Dsc::TypeHelpers.munge_boolean(value.to_s)
     end
 
     defaultto true
@@ -122,7 +123,7 @@ Puppet::Type.newtype(:dsc_<%= resource.friendlyname.downcase %>) do
     end
 <%    when property.bool? -%>
     munge do |value|
-      provider.munge_boolean(value.to_s)
+      PuppetX::Dsc::TypeHelpers.munge_boolean(value.to_s)
     end
 <%    when (property.int? || property.uint?) -%>
     munge do |value|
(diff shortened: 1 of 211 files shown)


Iristyle:
I have validated locally that all specs pass and that types that need to call `munge_boolean` can properly load the helpers.

Need a quick review from @ferventcoder and I think this can be merged in.

build/dsc/templates/dsc_type.rb.erb
```diff
@@ -1,4 +1,5 @@
 require 'pathname'
+require File.expand_path(File.join(File.dirname(__FILE__), '..', '..', 'puppet_x/puppetlabs/dsc_type_helpers'))
```
ferventcoder: Why doesn't this just follow the example on line 5?

build/dsc/templates/dsc_type.rb.erb
```diff
@@ -1,4 +1,5 @@
 require 'pathname'
+require File.expand_path(File.join(File.dirname(__FILE__), '../../puppet_x/puppetlabs/dsc_type_helpers'))
```
ferventcoder: Move to after line 5, follow what that line is doing.

ferventcoder:
@Iristyle if you can make that one adjustment, then rebuild all of the types, I'm +1 on merging this.

build/dsc/templates/dsc_type.rb.erb
Iristyle: Is there a compelling reason for `'../..' + 'puppet/_x/puppetlabs/dsc_type_helpers'` vs `'../../puppet/_x/puppetlabs/dsc_type_helpers'`?

In other words, is that extra string concat there for some good reason?

ferventcoder: Readability.

Iristyle: Also thought I was using standard module mechanism for `require` there, but looking at a few different projects, I see now that we don't really have a standard ;0

It's a hodgepdge of `File.expand_path`, `Pathname.new(__FILE__).dirname` and `require_relative`.

I actually think that `require_relative` is the best way of doing this that I know of - but maybe your firsthand experience says otherwise?

Iristyle:
@ferventcoder updated per your request

build/dsc/templates/dsc_type.rb.erb
ferventcoder: For `require_relative`, just haven't used it so I don't know what issues it may have or if it would just work without issue.

Iristyle: https://github.com/puppetlabs/puppetlabs-aws/search?utf8=%E2%9C%93&q=require_relative

build/dsc/templates/dsc_type.rb.erb
Iristyle: Yeah, oversight in my haste to get this shipped - thanks!

PR #103, opened 2015-09-22
(MODULES-2244) Evaluate Supported DSC Resources

jpogran:
We were including all DSC Resources from the PowerShell GitHub repo with
all of their files by default. This added more files than we wanted to
include in our distribution, as well as including DSC Resources we did not
want to include.

- [x] Implemented a blacklist of DSC Resources to be excluded from the vendored DSC Resources
- [x] Removed all test, example, and extraneous data files
- [x] Fix clean not being called properly on a clean/build workflow
- [x] Remove all extraneous files, keep license files


Iristyle:
Looks like the file count went up from 200-something to 468 - sweet.

I will try to review and merge this in more detail later today.

jpogran:
Removing non code files results in removing the DSC Resource for setting up a DSC Pull Server, xDscWebService and associated folders. Do we want to include this or remove it?

Iristyle:
@jpogran we'll do the update to new versions of the types in another PR - this PR was still dragging in stuff newer than dc5876d4cad9dce452106a79e2ab6e4f72c5b96b

Iristyle:
Superceded by #114

Iristyle:
WIP PR to update vendored resources at #115

PR #104, opened 2015-09-28
(WIP)(MODULES-2641) Notify on RebootRequired

jpogran:
Initiate a reboot when a DSC Resource requires a reboot after execution.

- [x] Detect `rebootrequired` property returned from DSC
- [ ] Add `autonotify` to all dsc types
- [x] Query the catalog to find `dsc_reboot` defined resource if present
- [x] Call method on `dsc_reboot`, if present, to initiate a reboot

*Note*: Tests are failing currently because `autonotify` on puppet < 4.0

### DIFF
--- a/build/dsc/templates/dsc_type.rb.erb
+++ b/build/dsc/templates/dsc_type.rb.erb
@@ -103,6 +103,10 @@ Puppet::Type.newtype(:dsc_<%= resource.friendlyname.downcase %>) do
     defaultto { :<%= begin resource.ensure_value rescue 'present' end %> }
   end
 
+if Puppet.version >= '4.0.0'
+  autonotify(:reboot) { ['dsc_reboot'] }
+end
+
 <%  resource.properties.each do |property| -%>
   # Name:         <%= property.name %>
   # Type:         <%= property.type %>
--- a/lib/puppet/provider/base_dsc/powershell.rb
+++ b/lib/puppet/provider/base_dsc/powershell.rb
@@ -57,8 +57,12 @@ EOT
     output = powershell(powershell_args, script_content)
     Puppet.debug "Create Dsc Resource returned: #{output}"
     data = JSON.parse(output)
+
     fail(data['errormessage']) if !data['errormessage'].empty?
-    true
+
+    notify_status if data['rebootrequired'] == true
+
+    data
   end
 
   def destroy
@@ -67,8 +71,56 @@ EOT
     output = powershell(powershell_args, script_content)
     Puppet.debug "Destroy Dsc Resource returned: #{output}"
     data = JSON.parse(output)
+
     fail(data['errormessage']) if !data['errormessage'].empty?
-    true
+
+    notify_status if data['rebootrequired'] == true
+
+    data
+  end
+
+  def notify_status
+    Puppet.info "A reboot is required to progress further. Notifying Puppet."
+
+    catalog_resource = resource.catalog.resources.find_all do |r|
+      r.is_a?(Puppet::Type.type(:reboot)) && r.name == 'dsc_reboot'
+    end
+
+    unless catalog_resource.any?
+      Puppet.warning "No reboot resource found in the graph that has 'dsc_reboot' as it's name. Cannot signal reboot to Puppet."
+      return
+    end
+
+    target_resource = resolve_resource(catalog_resource[0])
+
+    if target_resource.respond_to?(:reboot_required)
+      target_resource.reboot_required
+    else
+      Puppet.warning "Reboot resource does not have :reboot_required method implemented. Cannot signal reboot to Puppet."
+      return
+    end
+  end
+
+  def resolve_resource(reference)
+    if reference.is_a?(Puppet::Type)
+      # Probably from a unit test, use the resource as-is
+      return reference
+    end
+
+    if reference.is_a?(Puppet::Resource)
+      # Already part resolved - puppet apply?
+      # join it to the catalog where we live and ask it to resolve
+      reference.catalog = resource.catalog
+      return reference.resolve
+    end
+
+    if reference.is_a?(String)
+      # 3.3.0 catalogs you need to resolve like so
+      return resource.catalog.resource(reference)
+    end
+
+    # If we got here, panic
+    raise "Don't know how to convert '#{reference.inspect}' of class #{reference.class} into a resource"
   end
 
   def format_dsc_value(dsc_value)
--- a/lib/puppet/type/dsc_xcomputer.rb
+++ b/lib/puppet/type/dsc_xcomputer.rb
@@ -50,6 +50,10 @@ Puppet::Type.newtype(:dsc_xcomputer) do
     defaultto { :present }
   end
 
+if Puppet.version >= '4.0.0'
+  autonotify(:reboot) { ['dsc_reboot'] }
+end
+
   # Name:         Name
   # Type:         string
   # IsMandatory:  True
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -2,5 +2,6 @@ fixtures:
   forge_modules:
     stdlib: "puppetlabs/stdlib"
     powershell: "puppetlabs/powershell"
+    reboot: "puppetlabs/reboot"
   symlinks:
     "dsc": "#{source_dir}"
--- a/README.md
+++ b/README.md
@@ -58,6 +58,17 @@ dsc_windowsfeature {'IIS':
 
 All DSC Resource names and parameters have to be in lowercase, e.g: `dsc_windowsfeature` or `dsc_name`.
 
+### Handling Reboots with DSC
+
+You will need to add the following `reboot` resource to your manifest. It must have the name `dsc_reboot` for the `dsc` module to find and use it.
+
+~~~puppet
+reboot { 'dsc_reboot' :
+    message => 'DSC has requested a reboot',
+    when => 'refreshed'
+}
+~~~
+
 ### Installing Packages with DSC
 
 You can install MSIs or EXEs with DSC using the Puppet type `dsc_package` which maps to the `Package` DSC Resource.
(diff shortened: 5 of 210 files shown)


build/dsc/templates/dsc_type.rb.erb
```diff
@@ -98,11 +98,25 @@ Puppet::Type.newtype(:dsc_<%= resource.friendlyname.downcase %>) do
...
+    if last_dsc_result['rebootrequired'] == true
+      ['reboot_the_node']
+    end
+  }
+
   ensurable do
     newvalue(:exists?) { provider.exists? }
-    newvalue(:<%= begin resource.ensure_value rescue 'present' end %>) { provider.create }
+    newvalue(:<%= begin resource.ensure_value rescue 'present' end %>) {
+      last_dsc_result = sprovider.create
```
Iristyle: I don't think that you meant `sprovider` here

lib/puppet/type/dsc_xcomputer.rb
```diff
@@ -48,9 +48,23 @@
...
+  }
+
   ensurable do
-    newvalue(:exists?) { provider.exists? }
-    newvalue(:present) { provider.create }
+    newvalue(:exists?) { provider.exists }
+    newvalue(:present) {
+      last_dsc_result = provider.create
+      require 'pry';binding.pry;
+    }
```
Iristyle: Don't forget to return a `true` or `false` here based on success.

lib/puppet/type/dsc_xcomputer.rb
```diff
@@ -48,9 +48,23 @@
...
+    require 'pry';binding.pry;
+    if last_dsc_result['rebootrequired'] == true
+      ['reboot_the_node']
+    end
+  }
+
   ensurable do
-    newvalue(:exists?) { provider.exists? }
-    newvalue(:present) { provider.create }
+    newvalue(:exists?) { provider.exists }
```
Iristyle: I think we might as well capture the DSC return values every time.

lib/puppet/type/dsc_xcomputer.rb
```diff
@@ -48,9 +48,23 @@
   newparam(:name, :namevar => true ) do
   end
 
+  #newproperty(:last_dsc_result) do
+  #end
```
Iristyle: A plain old Ruby property `attr_reader :last_dsc_result` should be fine here.  I don't think we ever intend to expose these raw values to Puppet in any way.

lib/puppet/provider/base_dsc/powershell.rb
```diff
@@ -67,8 +68,9 @@ def destroy
     output = powershell(powershell_args, script_content)
     Puppet.debug "Destroy Dsc Resource returned: #{output}"
     data = JSON.parse(output)
+    Puppet.warning "A reboot is required to progress further. Please reboot the system." if data['rebootrequired'] == true
```
Iristyle: Notice vs Warning - thoughts?

ferventcoder: We talked about this one. The original one will send a warning, so we decided it would be appropriate here as well. Plus it may give notice that the user needs to do something. As a secondary step I was thinking about how we could remove the warning if the auto notified item was in the catalog.

lib/puppet/type/dsc_xcomputer.rb
```diff
@@ -48,9 +48,23 @@
...
   end
 
+  # ensurable do
+  #   newvalue(:exists?) { provider.exists? }
+  #   newvalue(:present) { provider.create }
+  #   defaultto { :present }
+  # end
+
+  attr_reader :last_dsc_result
+  autonotify(:reboot) { ['dsc_reboot'] }
```
Iristyle: So we still want to keep the autonotify.. but... I think we can remove the rest.

Do you want to start poking around at the code to retrieve the resource named `['dsc_reboot']` in the event that we get a `RebootRequired` back from `Invoke-DscResource`?

lib/puppet/provider/base_dsc/powershell.rb
```diff
@@ -67,10 +71,64 @@ def destroy
...
+      r.is_a?(Puppet::Type.type(:reboot))
+    end
+
+    require 'pry';binding.pry;
+    if catalog_resource.any?
+      return
+    end
+
+    target_resource = resolve_resource(catalog_resource[0])
+    require 'pry';binding.pry;
```
Iristyle: whoops ;0

metadata.json
```diff
@@ -44,6 +44,10 @@
     {
       "name": "puppetlabs-powershell",
       "version_requirement": ">= 1.0.1 < 2.0.0"
+    },
+    {
+      "name": "puppetlabs-reboot",
+      "version_requirement": ">= 1.0.1 < 2.0.0"
```
Iristyle: Do we want to require reboot as a dependency? Or do we want users to opt in?

lib/puppet/provider/base_dsc/powershell.rb
```diff
@@ -67,10 +71,64 @@ def destroy
...
+
+    name = target_resource.name
+    type = target_resource.type
+    resource_key = [type, name].join('/')
+
+    if target_resource.respond_to?(:autonotify)
+      target_resource.autonotify
+    end
```
Iristyle: I don't believe there's any need to call `autonotify` here...

lib/puppet/provider/base_dsc/powershell.rb
```diff
@@ -67,10 +71,64 @@ def destroy
...
+
     fail(data['errormessage']) if !data['errormessage'].empty?
-    true
+
+    initiate_reboot if data['rebootrequired'] == true
+
+    data
+  end
+
+  def initiate_reboot
```
Iristyle: You're not really initiating a reboot here.  You're setting a status on another resource, that will later perform the reboot.

lib/puppet/provider/base_dsc/powershell.rb
jpogran: This was a placeholder until I got the correct method to look for as well as call.

Iristyle: Got it - sorry for the mix-up there.  I think we want to use `:request_reboot` or `:reboot_desired` or similar.

Any thoughts on the name @ferventcoder ?

lib/puppet/provider/base_dsc/powershell.rb
```diff
@@ -67,10 +71,59 @@ def destroy
...
     fail(data['errormessage']) if !data['errormessage'].empty?
-    true
+
+    notify_status if data['rebootrequired'] == true
+
+    data
   end
 
+  def notify_status
+    Puppet.warning "A reboot is required to progress further. Notifying Puppet."
```
Iristyle: I think we ideally want 2 messages here... one that we were able to request the reboot.

Another when we couldn't find a matching resource in the graph to assist with triggering the reboot

lib/puppet/provider/base_dsc/powershell.rb
jpogran: I vote for `:reboot_required` or `:request_reboot`

metadata.json
Iristyle: Sorry @jpogran - I wasn't necessarily saying you had to remove it... thinking we could ask for a few more opinions on this, as there are some pros and cons to adding reboot as a dependency.

@ferventcoder / @cowofevil any thoughts?

build/dsc/templates/dsc_type.rb.erb
```diff
@@ -103,6 +103,8 @@ Puppet::Type.newtype(:dsc_<%= resource.friendlyname.downcase %>) do
     defaultto { :<%= begin resource.ensure_value rescue 'present' end %> }
   end
 
+  autonotify(:reboot) { ['dsc_reboot'] }
```
Iristyle: We need to do the Puppet version check here before using the autonotify.

We do a similar check in:
https://github.com/puppetlabs/puppetlabs-acl/blob/b1bd084fe7a43c99dbe44067a4898171b9fe6cf1/lib/puppet/util/monkey_patches.rb#L134

build/dsc/templates/dsc_type.rb.erb
```diff
@@ -103,6 +103,10 @@ Puppet::Type.newtype(:dsc_<%= resource.friendlyname.downcase %>) do
     defaultto { :<%= begin resource.ensure_value rescue 'present' end %> }
   end
 
+if Puppet.version >= '4.0.0'
```
Iristyle: So the problem with this is that we also want an edge to exist prior to Puppet 4.

Thinking about this a little more, it may be easier for us (given Puppet 3.x is the mix) to just write a little code similar to what's in here https://github.com/puppetlabs/puppet/blob/d2199af906fa5ca19ec0c37c82794934ea4fb4b5/lib/puppet/type.rb#L2111-L2143

The important part is mostly finding the resource and doing `reqs << Puppet::Relationship.new(self, dep)`

Then we can do the edge manually instead of using `autonotify` syntax.

lib/puppet/provider/base_dsc/powershell.rb
```diff
@@ -67,8 +71,67 @@ def destroy
...
+      # TODO: should this be a hard fail? probably?
+      Puppet.warning "Reboot resource 'dsc_reboot' is already connected to #{resource}"
+      return
+    end
+
+    # prevent addition of the same edge in the graph if a manual one exists
+    if !resource.catalog.relationship_graph.edge?(edge.source, edge.target)
+      edge.callback = :refresh
+      edge.event = :ALL_EVENTS
+      resource.catalog.relationship_graph.add_edge(edge)
```
jpogran: At this point we receive a message stating that our dsc_reboot resource has a negative number of dependencies. This comes from puppet relationshipgraph on line [73](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/graph/relationship_graph.rb#L73). I don't understand why this method does this, as `> 0' misses 0 as a non negative number.

Iristyle:
Superceded by #142

PR #105, opened 2015-09-28
(MODULES-2485) Remove LCM Pull RefreshMode

jpogran:
Remove the ability to set the LCM RefreshMode to Pull, as it will cause
errors during execution without setting more information than we
currently support in the dsc::lcm_config type. Since this type is no
longer needed in our module, this is not a loss.

### DIFF
--- a/spec/defines/lcm_config_spec.rb
+++ b/spec/defines/lcm_config_spec.rb
@@ -14,7 +14,7 @@ RSpec.describe 'dsc::lcm_config', :type => :define do
     }
   end
 
-  ['Disabled', 'Push', 'Pull'].each do |mode|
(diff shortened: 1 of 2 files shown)


spec/defines/lcm_config_spec.rb
```diff
@@ -31,7 +31,7 @@
     end
   end
 
-  ['disabled', 'push', 'pull', 'foo', 'bar'].each do |mode|
+  ['disabled', 'push', 'foo', 'bar'].each do |mode|
```
ferventcoder: Actually add this one back. And add 'Pull' to this as well.

ferventcoder:
:shipit: Once this is green it's good to go.

Iristyle:
:+1:

PR #107, opened 2015-09-30
(MODULES-2615) Prevent empty password for PSCredential

jpogran:
If an empty password is provided to any paramter that is a PSCredential
then we will fail.

Detect an empty password and warn the user it needs to be provided.

### DIFF
--- a/lib/puppet_x/puppetlabs/dsc_type_helpers.rb
+++ b/lib/puppet_x/puppetlabs/dsc_type_helpers.rb
@@ -58,6 +58,10 @@ module PuppetX
         unless extraneous.empty?
           fail "#{name} includes invalid keys: #{extraneous.join(',')}"
         end
+
+        if value.values.any?{|v| v.nil? || v.length == 0}
+          fail "Both User and Password must not be empty"
+        end
       end
 
       def self.validate_mof_type(mof_type, embeddedinstance_name, name, value)
--- a/build/dsc/templates/dsc_type_spec.rb.erb
+++ b/build/dsc/templates/dsc_type_spec.rb.erb
@@ -34,6 +34,12 @@ describe Puppet::Type.type(:dsc_<%= resource.friendlyname.downcase %>) do
   end
 <%  end # resource.ensurable? -%>
 <%  resource.properties.each do |property| -%>
+<%    if property.credential? -%>
+
+  it "should not accept empty password for dsc_<%= property.name.downcase %>" do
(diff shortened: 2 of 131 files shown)


lib/puppet_x/puppetlabs/dsc_type_helpers.rb
```diff
@@ -58,6 +58,10 @@ def self.validate_MSFT_Credential(name, value)
         unless extraneous.empty?
           fail "#{name} includes invalid keys: #{extraneous.join(',')}"
         end
+
+        if value.values.any?{|v| v.nil? || v.length == 0}
+          fail "Both User and Password must not be empty"
```
jpogran: Initially thought we should allow an empty password as PSCredential allows it. However cannot create a SecureString with a null or empty string, so would fail there anyway.

lib/puppet_x/puppetlabs/dsc_type_helpers.rb
```diff
@@ -58,6 +58,10 @@ def self.validate_MSFT_Credential(name, value)
         unless extraneous.empty?
           fail "#{name} includes invalid keys: #{extraneous.join(',')}"
         end
+
+        if value.values.any?{|v| v.nil? || v.length == 0}
```
jpogran: This doesn't tell the user which value is empty, just that there is something empty. Is there a better way that isn't hardcoded?

lib/puppet_x/puppetlabs/dsc_type_helpers.rb
ferventcoder: :+1:

lib/puppet_x/puppetlabs/dsc_type_helpers.rb
ferventcoder: That's fine. The message is appropriate.

ferventcoder:
Generate the rest. Looks good.

jpogran:
The rest of the specs are generated and pass

lib/puppet_x/puppetlabs/dsc_type_helpers.rb
Iristyle: Since we were already iterating over user and password a few lines above, why not just adjust the logic there rather than introduce new logic?

https://github.com/jpogran/puppetlabs-dsc/blob/20461a78cf95e64588a6eaa152824c798961ff59/lib/puppet_x/puppetlabs/dsc_type_helpers.rb#L44-L48

jpogran: Addressed in new PR since this one is closed

PR #109, opened 2015-10-01
(MODULES-2615) Handle Empty Passwords for PSCredentials

jpogran:
Handle checking if a username or password is not provided better by
checking in the first iteration and not a second time.

### DIFF
--- a/lib/puppet_x/puppetlabs/dsc_type_helpers.rb
+++ b/lib/puppet_x/puppetlabs/dsc_type_helpers.rb
@@ -44,6 +44,7 @@ module PuppetX
         required.each do |key|
           if value[key]
             fail "#{key} for #{name} should be a String" unless value[key].is_a? String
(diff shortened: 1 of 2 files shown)


Iristyle:
:+1: as long as you ran acceptance and it passed?

ferventcoder:
Also a thumbs up.

ferventcoder:
Implement on types.

jpogran:
@ferventcoder Types don't need to be updated because they look to see it raise an error, not for a string match. Should that be changed?

ferventcoder:
@jpogran oh, missed that. Super awesome!

Iristyle:
Ran through acceptance here at Puppet HQ - looks g2g.

Iristyle:
And yeah, as a point of note - the more shared code that lives outside of the generated types, the better IMHO.

PR #112, opened 2015-10-07
(MODULES-2270) refactor provider to add property accessors

Iristyle:


Iristyle:
@joshcooper a simple step 1 towards implementing prefetch... all specs pass, puppet apply works locally still.

PR #114, opened 2015-10-13
(MODULES-2244) Evaluate Supported DSC Resources

Iristyle:
Channeling @jpogran from the other PR:

We were including all DSC Resources from the PowerShell GitHub repo with
all of their files by default. This added more files than we wanted to
include in our distribution, as well as including DSC Resources we did not
want to include.

- [x] Implemented a blacklist of DSC Resources to be excluded from the vendored DSC Resources
- [x] Removed all test, example, and extraneous data files
- [x] Fix clean not being called properly on a clean/build workflow
- [x] Remove all extraneous files, keep license files


Iristyle:
Supercedes #103

Iristyle:
@jpogran / @ferventcoder this is ready for review now

build/dsc.rake
```diff
@@ -80,7 +94,9 @@ eod
     task :clean, [:dsc_resources_path] do |t, args|
       dsc_resources_path = args[:dsc_resources_path] || default_dsc_resources_path
       puts "Cleaning #{item_name}"
-      FileUtils.rm_rf "#{default_dsc_resources_path}"
+      FileUtils.rm_rf "#{dsc_resources_path}"
+      FileUtils.rm_rf "#{vendor_dsc_resources_path}"
```
Iristyle: @jpogran this line in particular was necessary to properly remove vendored resources prior to revendoring

jpogran:
LGTM

PR #116, opened 2015-10-19
(MODULES-2244) Additional blacklist excludes

Iristyle:
* Remove WaitFor* resources
* Fix a bug that wasn't excluding all samples due to case-sensitivity
* Abort type generation if WaitFor* resources have been accidentally re-included


ferventcoder:
This looks good. I'm not totally sold on the need for a rake abort if WaitFor* resources are found, but if a user trying to build their own custom items runs into that, I guess they will follow up with an issue.

build/dsc.rake
```diff
@@ -110,6 +110,8 @@ eod
     task :build, [:module_path] do |t, args|
       module_path = args[:module_path] || default_dsc_module_path
       m = Dsc::Manager.new
+      wait_for_resources = Dir["#{module_path}/**/*WaitFor*"]
+      fail "WaitFor* resources found - aborting type building!\n\n#{wait_for_resources}\n" if !wait_for_resources.empty?
```
ferventcoder: This is interesting...

Iristyle: Yeah, this is not an ideal way of doing this I agree.

It's mainly a guard against accidentally re-vendoring anything related to `WaitFor`.  The hope would be that we can replace our current manual vendoring process with something automated in the future, and remove the need for this altogether.

PR #118, opened 2015-10-23
(MODULES-2562) Handle uint arrays

jpogran:
Some DSC Resource parameters accept uint arrays. Account for this in the
type generation by detecting the unit array and adding correct
validation and spec generation

Note this relies on PR #119 being merged first in order to build the types correctly.

### DIFF
--- a/build/dsc/templates/dsc_type.rb.erb
+++ b/build/dsc/templates/dsc_type.rb.erb
@@ -168,9 +168,19 @@ Puppet::Type.newtype(:dsc_<%= resource.friendlyname.downcase %>) do
       PuppetX::Dsc::TypeHelpers.validate_<%= property.embeddedinstance_class_name %>(<%= !property.credential? ? 'mof_type_map, ' : '' %>"<%= property.name %>", value)
 <%      end -%>
 <%    when property.array? -%>
+<%      if property.int_array? -%>
+      unless value.kind_of?(Array) || value.kind_of?(Numeric) || value.to_i.to_s == value
+        fail("Invalid value '#{value}'. Should be an integer or an array of integers")
+      end
+<%      elsif property.uint_array? -%>
+      unless value.kind_of?(Array) || (value.kind_of?(Numeric) && value >= 0) || (value.to_i.to_s == value && value.to_i >= 0)
+        fail("Invalid value '#{value}'. Should be an integer or an array of integers")
+      end
+<%      else -%>
       unless value.kind_of?(Array) || value.kind_of?(String)
         fail("Invalid value '#{value}'. Should be a string or an array of strings")
       end
+<%      end -%>
 <%    when property.bool? -%>
 <%    when property.int? -%>
       unless value.kind_of?(Numeric) || value.to_i.to_s == value
@@ -215,6 +225,9 @@ Puppet::Type.newtype(:dsc_<%= resource.friendlyname.downcase %>) do
       value.kind_of?(Hash) ?
         [PuppetX::Dsc::TypeHelpers.munge_embeddedinstance(mof_type_map, value)] :
         value.map { |v| PuppetX::Dsc::TypeHelpers.munge_embeddedinstance(mof_type_map, v) }
+<%     elsif (property.int_array? || property.uint_array?) -%>
+      v = PuppetX::Dsc::TypeHelpers.munge_integer(value)
+      v.is_a?(Array) ? v : Array(v)
 <%      else -%>
       Array(value)
 <%      end -%>
@@ -229,7 +242,7 @@ Puppet::Type.newtype(:dsc_<%= resource.friendlyname.downcase %>) do
     end
 <%    when (property.int? || property.uint?) -%>
     munge do |value|
-      value.to_i
+      PuppetX::Dsc::TypeHelpers.munge_integer(value)
     end
 <%    else -%>
 <%    end -%>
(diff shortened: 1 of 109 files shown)


jpogran:
TODO: Have to change the dsc_type_spec.rb.erb to allow single uint values to be passed

build/dsc/templates/dsc_type_spec.rb.erb
```diff
@@ -162,6 +162,12 @@ describe Puppet::Type.type(:dsc_<%= resource.friendlyname.downcase %>) do
   end
 <%      end #  @spec_test_values['munged_ints']... -%>
 
+<%    elsif property.int_array? || property.int? || property.uint? || property.uint_array?  # if property.int? -%>
```
jpogran: This is the only way I could short circuit the existing int/uint logic. Int and uint arrays should accept single values. but can't accept a string. Is there a better way?

Iristyle:
Rebase / update commit messages, and I'll take a look.

build/dsc/templates/dsc_type.rb.erb
```diff
@@ -168,9 +168,19 @@ Puppet::Type.newtype(:dsc_<%= resource.friendlyname.downcase %>) do
       PuppetX::Dsc::TypeHelpers.validate_<%= property.embeddedinstance_class_name %>(<%= !property.credential? ? 'mof_type_map, ' : '' %>"<%= property.name %>", value)
 <%      end -%>
 <%    when property.array? -%>
+<%      if property.int_array? -%>
+      unless value.kind_of?(Array) || value.kind_of?(Numeric) || value.to_i.to_s == value || value.to_i >= 0
```
jpogran: @Iristyle this more along the lines you were thinking?

build/dsc/templates/dsc_type.rb.erb
```diff
@@ -229,7 +242,8 @@ Puppet::Type.newtype(:dsc_<%= resource.friendlyname.downcase %>) do
     end
 <%    when (property.int? || property.uint?) -%>
     munge do |value|
-      value.to_i
+      v = PuppetX::Dsc::TypeHelpers.munge_integer(value)
```
jpogran: can this be made shorter?

Iristyle: Sadly, no..

build/dsc/templates/dsc_type.rb.erb
Iristyle: Do we want `value.to_i >= 0` check on a signed int?

build/dsc/templates/dsc_type.rb.erb
```diff
@@ -229,7 +242,8 @@ Puppet::Type.newtype(:dsc_<%= resource.friendlyname.downcase %>) do
...
+      v.is_a?(Array) ? v : Array(v)
```
jpogran: this line is wrong. should not return an array for a single int/uint

Iristyle:
After some more discovery, we discovered some additional issues and opened https://tickets.puppetlabs.com/browse/MODULES-2759, which I think we want to mix into the fold first, at which point we can return to this PR.

build/dsc/templates/dsc_type_spec.rb.erb
Iristyle: They should still accept a string for Puppet 3.x.

I'll make sure I review everything carefully prior to merge.

build/dsc/templates/dsc_type.rb.erb
```diff
@@ -215,6 +225,9 @@ Puppet::Type.newtype(:dsc_<%= resource.friendlyname.downcase %>) do
       value.kind_of?(Hash) ?
         [PuppetX::Dsc::TypeHelpers.munge_embeddedinstance(mof_type_map, value)] :
         value.map { |v| PuppetX::Dsc::TypeHelpers.munge_embeddedinstance(mof_type_map, v) }
+<%     elsif (property.int_array? || property.uint_array?) -%>
```
Iristyle: This can never happen.  When dealing with embedded instances, the corresponding MOF type is always either `string` or `string[]`

spec/unit/puppet/type/dsc_package_spec.rb
```diff
@@ -202,12 +202,14 @@
...
+  it 'should accept int for dsc_returncode' do
+    dsc_package[:dsc_returncode] = [32, 64, 128]
+    expect(dsc_package[:dsc_returncode]).to eq([32, 64, 128])
   end
 
-  it 'should not accept uint for dsc_returncode' do
-    expect{dsc_package[:dsc_returncode] = 16}.to raise_error(Puppet::ResourceError)
+  it 'should accept uint for dsc_returncode' do
+    dsc_package[:dsc_returncode] = [32, 64, 128]
+    expect(dsc_package[:dsc_returncode]).to eq([32, 64, 128])
```
Iristyle: Something doesn't seem right here.  It should accept `int` and it should accept `uint`?

The MOF reads like:

```
[write] uint32 ReturnCode[];
```

I think there's a bug with the test generators.

Iristyle: Also keep in mind when we move over to the validation code in the helpers that there are ranges of acceptable values here (i.e. signed values have to reserve 1 bit for the sign)... so it's important to make sure tests against `int` reference `int`, etc.

Probably more cleanup to do in that subsequent PR.

Iristyle:
@jpogran you can check the commit log to see how I restructured / what I fixed

PR #120, opened 2015-10-29
(MODULES-2579) Handle Signed Integers

jpogran:
- [x] Fix signed integer validation
- [x] Add sint8[], sint16[], sint32[], and sint64[] to the parameter code generation
- [x] Add sint8[], sint16[], sint32[], and sint64[] to the validation helpers

*Note*: PR #118 fixes generating correct validation methods for integer array (uint32[]..etc) parameters. This PR doesn't try to fix that since it's already done.

### DIFF
--- a/build/dsc/property.rb
+++ b/build/dsc/property.rb
@@ -99,11 +99,11 @@ module Dsc
     end
 
     def int?
-      ["int16","int32","int64","sint16","sint32","sint64"].include?(type)
+      ["int8","int16","int32","int64","sint8","sint16","sint32","sint64"].include?(type)
     end
 
     def int_array?
-      ["int16[]","int32[]","int64[]"].include?(type)
+      ["int8[]","int16[]","int32[]","int64[]","sint8[]","sint16[]","sint32[]","sint64[]"].include?(type)
     end
 
     def string?
--- a/build/dsc/templates/dsc_type.rb.erb
+++ b/build/dsc/templates/dsc_type.rb.erb
@@ -173,7 +173,7 @@ Puppet::Type.newtype(:dsc_<%= resource.friendlyname.downcase %>) do
       end
 <%    when property.bool? -%>
 <%    when property.int? -%>
-      unless value.kind_of?(Numeric) || value.to_i.to_s == value || value.to_i >= 0
+      unless value.kind_of?(Numeric) || value.to_i.to_s == value
           fail("Invalid value #{value}. Should be a signed Integer")
       end
 <%    when property.uint? -%>
--- a/lib/puppet/type/dsc_windowsprocess.rb
+++ b/lib/puppet/type/dsc_windowsprocess.rb
@@ -226,7 +226,7 @@ Puppet::Type.newtype(:dsc_windowsprocess) do
     def mof_type; 'sint32' end
     def mof_is_embedded?; false end
     validate do |value|
-      unless value.kind_of?(Numeric) || value.to_i.to_s == value || value.to_i >= 0
+      unless value.kind_of?(Numeric) || value.to_i.to_s == value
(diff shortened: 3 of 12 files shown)


Iristyle:
Looking good - I think you squash db65e5c / 2f8989b together.

Are there any affected tests?  Should probably be an update there to demonstrate why the new change works if possible.

jpogran:
Will squash and re-push those.

We don't see a change to any tests because we have no current types that have the missing values. Forcing a parameter manually to a value like 'sint8' produces the expected change, but since we don't have the types we don't get anything right now. We will see more changes when this PR and PR #118 get merged together.

Iristyle:
Ok @jpogran ... so to understand what I meant about the tests, you have to kind of debug / sniff around a little.  We know based on your changes in 84e81c2 that we have some `sint` types where we've affected the type generation now.  So this means that:

- There were never any tests that tried to pass negative values to any of the affected validation checks (in which case we need to add some)
- There were tests, but they were errantly passing / the check we removed was ineffectual

So digging in a little more, I see that we *did* have tests that tried to pass negative values.  Take a look at these for example - https://github.com/puppetlabs/puppetlabs-dsc/blob/master/spec/unit/puppet/type/dsc_xblautobitlocker_spec.rb#L91-L124

So this means tests were passing in spite of the wrong check (which ends up being superfluous in this case)... and in turn that makes me distrustful of the logic employed in the first place for integer validation that's being used.


Here are the individual checks for `-32`:

```
[7] pry(#<Puppet::Type::Dsc_xblautobitlocker::ParameterDsc_mindiskcapacitygb>)> value = -32
=> -32
[8] pry(#<Puppet::Type::Dsc_xblautobitlocker::ParameterDsc_mindiskcapacitygb>)> value.kind_of?(Numeric)
=> true
[9] pry(#<Puppet::Type::Dsc_xblautobitlocker::ParameterDsc_mindiskcapacitygb>)> value.to_i.to_s == value
=> false
[10] pry(#<Puppet::Type::Dsc_xblautobitlocker::ParameterDsc_mindiskcapacitygb>)> value.to_i >= 0
=> false
```

And for `"-32"`:

```
[11] pry(#<Puppet::Type::Dsc_xblautobitlocker::ParameterDsc_mindiskcapacitygb>)> value = "-32"
=> "-32"
[12] pry(#<Puppet::Type::Dsc_xblautobitlocker::ParameterDsc_mindiskcapacitygb>)> value.kind_of?(Numeric)
=> false
[13] pry(#<Puppet::Type::Dsc_xblautobitlocker::ParameterDsc_mindiskcapacitygb>)> value.to_i.to_s == value
=> true
[14] pry(#<Puppet::Type::Dsc_xblautobitlocker::ParameterDsc_mindiskcapacitygb>)> value.to_i >= 0
=> false
```

When used with `unless`, we can see why the previous tests pass, because you only need a single `true` to prevent failing validation.  So basically the `>= 0` check was just a vestigial appendage and did no harm here (aside from causing confusion).  Looking at the integer checks, at least they're written as `unless (value.kind_of?(Numeric) && value >= 0) || (value.to_i.to_s == value && value.to_i >= 0)` 

So hopefully that explains what I was asking about - it's all a little confusing (partially due to the `unless`).  We should probably just rewrite the checks as an `if`, but let's hold off on that for now, given we're going to leverage the `validate_mof_type` helper soon-ish at https://github.com/puppetlabs/puppetlabs-dsc/blob/master/lib/puppet_x/puppetlabs/dsc_type_helpers.rb#L73-L99 which is more comprehensive a test anyhow.

So in summary, I think this is :+1:  but let's update the first commit message with a bit of info on why no tests required changing - summarizing the above.

Iristyle:
FYI - updated commit message, and fixed some omissions in dfbd3c0 ... merged accordingly

PR #123, opened 2015-10-30
(MODULES-2771) Ping master_manipulator Version

cowofevil:
The QA-2147 bug causes the DSC pipeline to break on PE 3.8.x. To work around
this issue temporarily we need to pin the version to known working release.
This fix should allow the acceptance tests to pass in the CI pipeline.
[skip-ci]

### DIFF
--- a/.sync.yml
+++ b/.sync.yml
@@ -20,7 +20,7 @@ Gemfile:
     ':system_tests':
       - gem: beaker
       - gem: master_manipulator
-        version: '~> 1.1'
+        version: '1.1.2'
     ':build':
       - gem: 'librarian-repo'
         git: 'https://github.com/msutter/librarian-repo.git'


Iristyle:
So looks like we're way out of date with what modulesync does - I ran and it produced this commit from this change -> https://github.com/Iristyle/puppetlabs-dsc/commit/351be2e70326e3444d118aa0e1a88fecbcb9b676

That's way more than I wanted to merge / figure out right now, so I instead grabbed the relevant changes for this gem, manually updated `Gemfile` and merged to (hopefully) unblock CI.

Iristyle:
@ferventcoder if you want to take a look a the `modulesync` problems next sprint, that would be :+1:

ferventcoder:
I saw your commit @Iristyle, but it looks like you were pointed to the wrong modulesync_config? https://github.com/puppetlabs/modulesync_configs/tree/windows

Iristyle:
Yes, thanks @ferventcoder ... I re-ran from `windows` branch and the commit looks better now, but still has other extraneous bits:

https://github.com/Iristyle/puppetlabs-dsc/commit/57ade7c1764ebff3695e13b8b902ae3e3f3c13d1

PR #128, opened 2015-11-04
(MODULES-2668) Document Windows 2003 LFN issues

Iristyle:
- Suggest users employ environments to prevent installs on 2003

### DIFF
--- a/README.md
+++ b/README.md
@@ -328,6 +328,8 @@ For Puppet 4.2.2+ (and 3.8.2) we've lessened the possibility of the issue occurr
 
 If you are affected by this issue, a known workaround is to download the `.tar.gz` from the forge and use `puppet module install` using the file rather than directly from the forge.
 
+- Windows 2003 is not supported.  Further, when installing the module on a Puppet master to the default `production` environment, this module ***will*** cause pluginsync to fail on Windows 2003 agents due to an issue with [LFN (long file names)](https://tickets.puppetlabs.com/browse/PUP-4866).  To workaround this, use [Puppet environments](https://docs.puppetlabs.com/puppet/latest/reference/environments.html) to create an environment separate from `production`, without the DSC module installed, to host Windows 2003 nodes.
+
 ### Running Puppet and DSC without Administrative Privileges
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -328,6 +328,8 @@ For Puppet 4.2.2+ (and 3.8.2) we've lessened the possibility of the issue occurr
 
 If you are affected by this issue, a known workaround is to download the `.tar.gz` from the forge and use `puppet module install` using the file rather than directly from the forge.
 
+- Windows 2003 is not supported.  If installing the module to the default `production` environment on a Puppet master, pluginsync will attempt to install on Windows 2003 nodes within the `production` environment.  This will likely cause pluginsync failures, where the LFN issue is even more problematic (due to a longer user home directory). In this situation, [Puppet environments](https://docs.puppetlabs.com/puppet/latest/reference/environments.html) may be used to segregate nodes and control module installs.  Since modules may be configured independently per environment, this allows for the creation of an environment to isolate Windows 2003, and ensure the DSC module is not installed.
+
```
ferventcoder: This is good, but it doesn't really explain that the module will be synced down to Windows 2003 agents automatically and will cause issues. 

"If you have any Windows 2003 agents, this module will cause pluginsync to fail on those agents due to an issue with [LFN (Long File Name)](https://tickets.puppetlabs.com/browse/PUP-4866). As a workaround, consider putting your Windows 2003 agents in a separate environment isolated from your normal `production` environment, see [Puppet environments](https://docs.puppetlabs.com/puppet/latest/reference/environments.html)."

Iristyle: `Windows 2003 is not supported` is the first thing there, so that seems pretty early ;0

ferventcoder: @Iristyle I get that. For the uneducated about what the issue is, they will read that line and do the same thing I did, which is "Duh, I knew that already. So what's the problem?"

ferventcoder: Where the statement I'm providing is leading them down a path of "Oh...".

jpogran:
LGTM

PR #132, opened 2015-11-04
(maint) Update modulesync

Iristyle:


ferventcoder:
Once appveyor is happy, I'm happy. Did any of the other repositories get changes as well?

Iristyle:
I was only concerned with DSC atm

.gitignore
```diff
@@ -1,6 +1,7 @@
 pkg/
 Gemfile.lock
 Gemfile.local
+vendor/
```
Iristyle: This was previously removed in https://github.com/puppetlabs/puppetlabs-dsc/blob/25240bc5b4eae6cd9cc195856d012219053d762b/.gitignore ... and it looks like we'd have to modify the template to enact any change here: https://github.com/puppetlabs/modulesync_configs/blob/windows/moduleroot/.gitignore

ferventcoder: We'll need to adjust that one back out. I'll take care of that.

PR #133, opened 2015-11-04
(MODULES-2562) Fix Intermittent Failing Package Tests

cowofevil:
Some of the acceptance tests for the "dsc_package" resource were failing
sometimes. The root cause is believed to be a poor uninstaller for the NSI
test program package. The affected tests have been updated to brute force
uninstall the test package. Also, the test installer binary has been moved
into the repo for convenience.
[skip ci]


ferventcoder:
`\o/` Hoping this clears CI!

PR #134, opened 2015-11-04
(maint) update modulesync files

ferventcoder:
~~This should come after we pull in https://github.com/puppetlabs/puppetlabs-dsc/pull/132~~

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,6 @@
 pkg/
 Gemfile.lock
 Gemfile.local
-vendor/
 spec/fixtures/
 log/
 junit/
@@ -21,3 +20,4 @@ import
 .rbenv-gemsets
 .ruby-version
 *.lock
+vendor/
--- a/Gemfile
+++ b/Gemfile
@@ -25,12 +25,24 @@ def location_for(place_or_version, fake_version = nil)
   end
 end
 
+# The following gems are not included by default as they require DevKit on Windows.
+# You should probably include them in a Gemfile.local or a ~/.gemfile
+#gem 'pry' #this may already be included in the gemfile
+#gem 'pry-stack_explorer', :require => false
+#if RUBY_VERSION =~ /^2/
+#  gem 'pry-byebug'
+#else
+#  gem 'pry-debugger'
+#end
+
 group :development do
-  gem 'rspec', '~>2.14.1',                   :require => false
+  gem 'rake', '~>10.1',                      :require => false
+  gem 'rspec', '~>3.0',                      :require => false
   gem 'puppet-lint',                         :require => false
   gem 'puppetlabs_spec_helper', '~>0.10.3',  :require => false
   gem 'puppet_facts',                        :require => false
   gem 'mocha', '~>0.10.5',                   :require => false
+  gem 'pry',                                 :require => false
   gem 'puppet-blacksmith',                   :require => false
   gem 'mime-types', '~>2.99',                :require => false
 end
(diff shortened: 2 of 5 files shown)


ferventcoder:
It depends on some things that will need to be rebased.

Gemfile
```diff
@@ -40,6 +52,7 @@ group :system_tests do
 end
 
 group :build do
+  gem 'librarian-repo',     :require => false, :git => 'https://github.com/msutter/librarian-repo.git'
```
ferventcoder: Not sure how this came back.

appveyor.yml
```diff
@@ -26,7 +26,6 @@ matrix:
   fast_finish: true
 install:
 - SET PATH=C:\Ruby%RUBY_VER%\bin;%PATH%
-- gem install bundler --quiet --no-ri --no-rdoc
```
ferventcoder: Bye gem install bundler.

Gemfile
```diff
@@ -25,12 +25,24 @@ def location_for(place_or_version, fake_version = nil)
...
+#gem 'pry-stack_explorer', :require => false
+#if RUBY_VERSION =~ /^2/
+#  gem 'pry-byebug'
+#else
+#  gem 'pry-debugger'
+#end
+
 group :development do
+  gem 'rake',                                :require => false
   gem 'rspec', '~>2.14.1',                   :require => false
```
ferventcoder: This manages its own rspec. I am not sure if this is borked or not.

Gemfile
```diff
@@ -25,12 +25,24 @@ def location_for(place_or_version, fake_version = nil)
...
+#if RUBY_VERSION =~ /^2/
+#  gem 'pry-byebug'
+#else
+#  gem 'pry-debugger'
+#end
+
 group :development do
-  gem 'rspec', '~>2.14.1',                   :require => false
+  gem 'rake',                                :require => false
+  gem 'rspec', '~>3.0',                      :require => false
```
glennsarti: Is the rspec 3.0 ok as we can only use Ruby 1.9.3 for DSC MOF parsing?

glennsarti:
Do we need to squash the commits here?

ferventcoder:
Looks like it.

Gemfile
ferventcoder: You mean will it cause issues? Unknown.

ferventcoder:
well that went peachy :/

.gitignore
```diff
@@ -1,7 +1,6 @@
 pkg/
 Gemfile.lock
 Gemfile.local
-vendor/
```
glennsarti: Why is vendor being removed from gitignore?

ferventcoder: We may need to keep this one?

glennsarti: As per recommendation, users are encouraged to add their DSC modules `/build/vendor/custom`

https://github.com/puppetlabs/puppetlabs-dsc/blob/master/README_BUILD.md#getting-started

KosherSalt: My team had trouble with vendor/ being in .gitignore and trying to use /build/vendor/custom for our custom modules. I had to add the following line under vendor to our .gitignore file or it wouldn't push our custom modules to the remote repo.
...
vendor/
!/build/vendor
...

ferventcoder: @KosherSalt I think we have a different ticket to evaluate that. I agree that it is something we need to fix, but it probably won't be part of this current PR.

glennsarti: @KosherSalt https://tickets.puppetlabs.com/browse/MODULES-3042 is for your issue.

KosherSalt: Gotcha. I was more trying to say that having vendor in .gitignore impacted us when we tried to use the recommended /build/vendor/custom, so I didn't think it would impact that process to remove it completely.

glennsarti: Agreed.  I want to test that we don't create any temp files in there and may get checked in by accident.

PR #135, opened 2015-11-06
(MODULES-2520) Add link to Types document

jpogran:
Add a reference and link to the types document that is automatically
generated when building the dsc_* types.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,12 +15,13 @@
   * [Adding or Removing Windows Features](#adding-or-removing-windows-features)
   * [Website Installation Example](#website-installation-example)
   * [Optionally Configure the DSC LCM RefreshMode](#optionally-configure-the-dsc-lcm-refreshmode)
-5. [Limitations](#limitations)
+5. [Reference - An under-the-hood peek at what the module is doing and how](#reference)
+6. [Limitations](#limitations)
   * [Known Issues](#known-issues)
   * [Running Puppet and DSC without Administrative Privileges](#running-puppet-and-dsc-without-administrative-privileges)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -15,12 +15,13 @@
   * [Adding or Removing Windows Features](#adding-or-removing-windows-features)
   * [Website Installation Example](#website-installation-example)
   * [Optionally Configure the DSC LCM RefreshMode](#optionally-configure-the-dsc-lcm-refreshmode)
-5. [Limitations](#limitations)
+5. [Reference - An under-the-hood peek at what the module is doing and how](#reference)
```
jpogran: Copied this to aline with other module documentation, but I don't know that it's valid given this section does not explain 'what the module is doing and how'

ferventcoder: Looks great.

README.md
```diff
@@ -297,6 +298,12 @@ dsc::lcm_config {'disable_lcm':
 }
 ~~~
 
+## Reference
+
+### Types
+
+A comprehensive listing of all types included in the dsc module is available in the [types document](https://github.com/puppetlabs/puppetlabs-dsc/blob/master/types.md).
+
```
ferventcoder: The docs review will overhaul this if they don't find it valid.

README.md
cowofevil: This looks weird. I think it is more intuitive to have the TOC just have "Reference" without the extra annotation.

ferventcoder: It's kind of a starter. I imagine a docs review will start to fill out this section some more.

PR #140, opened 2015-11-09
(MODULES-2520) List all DSC Types in Table Format

jpogran:
List all the generated DSC types in a markdown table format that is
easier to read. Also, group by DSC Resource Module so its easier to find
resources by the module they are in.

### DIFF
--- a/build/dsc/manager.rb
+++ b/build/dsc/manager.rb
@@ -168,13 +168,23 @@ module Dsc
         file.write("For any system this module is installed on, use\n`Puppet describe typename` for more information.\n\n")
 
         file.write("#### WMF Core Types\n\n")
-        dsc_types.select { |t| t.match(/dsc_[^x]/)}.each do |dsc_type|
-          file.write("* #{dsc_type}\n")
+        file.write("Puppet Type | DSC Resource\n")
+        file.write("----------- | -----------------\n")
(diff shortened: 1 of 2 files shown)


build/dsc/manager.rb
```diff
@@ -168,13 +168,21 @@ def document_types(markdown_file_path, dsc_types)
...
 
         file.write("\n#### Community x Prefixed types\n\n")
-        dsc_types.select { |t| t.match(/dsc_x/)}.each do |dsc_type|
-          file.write("* #{dsc_type}\n")
+        resources.select { |t| t.instance_name.downcase.match(/dsc_x/)}.group_by { |dt| dt.ps_module.name}.each do |dsc_type|
+          file.write("##### #{dsc_type[0]}\n\n")
+          file.write("Puppet Type | DSC Resource Name\n")
+          file.write("----------- | -----------------\n")
+          dsc_type[1].each do | sb |
+            file.write("dsc_#{sb.friendlyname.downcase} | #{sb.friendlyname}\n")
```
jpogran: I chose this order to show Puppet types first, but should this be DSC Resource then Puppet type?

Iristyle:
This will be a total nerd-snipe, but you should be able to pretty easily generate hyperlinks for the DSC types ;0

ferventcoder:
This looks great!

PR #142, opened 2015-11-13
(MODULES-2641) Invoke-DscResource triggers reboot

Iristyle:
When a DSC resource indicates that a reboot is required via its
  RebootRequired property, notify the Puppet resource reboot named
  dsc_reboot if it exists, so that a reboot may be executed on demand.

 - Typically, edges in the graphs between each DSC resource and the
   Puppet DSC resources could be created through Puppet's autotnotify
   helper.  However, this wasn't added until Puppet 4 and this module
   supports Puppet 3.8, so a different approach is required.

 - Override the builddepends method of each DSC type, so that a Puppet
   relationship dsc_type['name'] => reboot['dsc_reboot'] may be
   created, when the dsc_reboot instance exists in the catalog.  This
   allows refresh events to be propagated to the dsc_reboot instance
   when the DSC PowerShell provider calls create.

 - To ensure that reboots only happen when Invoke-DscResource returns
   RebootRequired, set an attribute on the Windows reboot provider
   called reboot_required.  The dsc_reboot instance must be configured
   to only reboot when 'pending', which will now check this internal
   state in addition to the other checks it normally performs, like
   checking registry settings - pending file renames, computer rename,
   Windows Update, etc.

fixup with prior commit - regenerate types as well

### DIFF
--- a/lib/puppet/provider/base_dsc/powershell.rb
+++ b/lib/puppet/provider/base_dsc/powershell.rb
@@ -57,8 +57,12 @@ EOT
     output = powershell(powershell_args, script_content)
     Puppet.debug "Create Dsc Resource returned: #{output}"
     data = JSON.parse(output)
+
     fail(data['errormessage']) if !data['errormessage'].empty?
-    true
+
+    notify_reboot_pending if data['rebootrequired'] == true
+
+    data
   end
 
   def destroy
@@ -67,8 +71,30 @@ EOT
     output = powershell(powershell_args, script_content)
     Puppet.debug "Destroy Dsc Resource returned: #{output}"
     data = JSON.parse(output)
+
     fail(data['errormessage']) if !data['errormessage'].empty?
-    true
+
+    notify_reboot_pending if data['rebootrequired'] == true
+
+    data
+  end
+
+  def notify_reboot_pending
+    Puppet.info "A reboot is required to progress further. Notifying Puppet."
+
+    reboot_resource = resource.catalog.resource(:reboot, 'dsc_reboot')
+    unless reboot_resource
+      Puppet.warning "No reboot resource found in the graph that has 'dsc_reboot' as its name. Cannot signal reboot to Puppet."
+      return
+    end
+
+    if reboot_resource.provider.respond_to?(:reboot_required)
+      # internal API used to let reboot resource knows a reboot is pending
+      reboot_resource.provider.reboot_required = true
+    else
+      Puppet.warning "Reboot module must be updated, since resource does not have :reboot_required method implemented. Cannot signal reboot to Puppet."
+      return
+    end
   end
 
   def format_dsc_value(dsc_value)
--- a/lib/puppet_x/puppetlabs/dsc_type_helpers.rb
+++ b/lib/puppet_x/puppetlabs/dsc_type_helpers.rb
@@ -130,6 +130,35 @@ module PuppetX
           end
         end
       end
+
+      def self.should_add_reboot_relationship(resource, reboot_resource, pending_relationships)
+        return false if !reboot_resource
+
+        # edge exists in graph from previous resource, so don't add
+        return false if resource.catalog.relationship_graph.edge?(resource, reboot_resource) ||
+          # newly formed edges not yet in catalog include the edge, so don't add
+          pending_relationships.any? { |e| e.source == resource && e.target == reboot_resource }
+
+        true
+      end
+
+      # if an edge already exists from Reboot[dsc_reboot] to this resource
+      # Puppet will recognize the cyclic dependency automatically and fail with:
+      # Error: Failed to apply catalog: Found 1 dependency cycle:
+      # (Dsc_file[foo] => Reboot[dsc_reboot] => Dsc_file[foo])
+      def self.ensure_reboot_relationship(resource, pending_relationships)
+        reboot_resource = resource.catalog.resource(:reboot, 'dsc_reboot')
+
+        # do nothing if no Reboot[dsc_reboot] or already an edge from resource to it
+        if should_add_reboot_relationship(resource, reboot_resource, pending_relationships)
+          # otherwise build resource[name] => Reboot[dsc_reboot]
+          edge = Puppet::Relationship.new(resource, reboot_resource,
+              { :callback => :refresh, :event => :ALL_EVENTS })
+          pending_relationships << edge
+        end
+
+        pending_relationships
+      end
     end
   end
 end
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -2,5 +2,6 @@ fixtures:
   forge_modules:
     stdlib: "puppetlabs/stdlib"
     powershell: "puppetlabs/powershell"
+    reboot: "puppetlabs/reboot"
   symlinks:
     "dsc": "#{source_dir}"
--- a/README.md
+++ b/README.md
@@ -60,6 +60,17 @@ dsc_windowsfeature {'IIS':
 
 All DSC Resource names and parameters have to be in lowercase, e.g: `dsc_windowsfeature` or `dsc_name`.
 
+### Handling Reboots with DSC
+
+You will need to add the following `reboot` resource to your manifest. It must have the name `dsc_reboot` for the `dsc` module to find and use it.
+
+~~~puppet
+reboot { 'dsc_reboot' :
+    message => 'DSC has requested a reboot',
+    when => 'pending'
+}
+~~~
+
 ### Installing Packages with DSC
 
 You can install MSIs or EXEs with DSC using the Puppet type `dsc_package` which maps to the `Package` DSC Resource.
--- a/build/dsc/templates/dsc_type.rb.erb
+++ b/build/dsc/templates/dsc_type.rb.erb
@@ -237,6 +237,10 @@ Puppet::Type.newtype(:dsc_<%= resource.friendlyname.downcase %>) do
   end
 
 <%  end %>
+  def builddepends
+    pending_relations = super()
+    PuppetX::Dsc::TypeHelpers.ensure_reboot_relationship(self, pending_relations)
+  end
 end
 
 Puppet::Type.type(:dsc_<%= resource.friendlyname.downcase %>).provide :powershell, :parent => Puppet::Type.type(:base_dsc).provider(:powershell) do
--- a/lib/puppet/type/dsc_archive.rb
+++ b/lib/puppet/type/dsc_archive.rb
@@ -157,6 +157,10 @@ Puppet::Type.newtype(:dsc_archive) do
   end
 
 
+  def builddepends
+    pending_relations = super()
+    PuppetX::Dsc::TypeHelpers.ensure_reboot_relationship(self, pending_relations)
+  end
 end
 
 Puppet::Type.type(:dsc_archive).provide :powershell, :parent => Puppet::Type.type(:base_dsc).provider(:powershell) do
--- a/lib/puppet/type/dsc_environment.rb
+++ b/lib/puppet/type/dsc_environment.rb
@@ -105,6 +105,10 @@ Puppet::Type.newtype(:dsc_environment) do
   end
 
 
+  def builddepends
+    pending_relations = super()
+    PuppetX::Dsc::TypeHelpers.ensure_reboot_relationship(self, pending_relations)
+  end
 end
 
 Puppet::Type.type(:dsc_environment).provide :powershell, :parent => Puppet::Type.type(:base_dsc).provider(:powershell) do
--- a/lib/puppet/type/dsc_file.rb
+++ b/lib/puppet/type/dsc_file.rb
@@ -266,6 +266,10 @@ Puppet::Type.newtype(:dsc_file) do
   end
 
 
+  def builddepends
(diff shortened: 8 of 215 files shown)


lib/puppet_x/puppetlabs/dsc_type_helpers.rb
```diff
@@ -126,6 +126,24 @@ def self.validate_mof_type(mof_type, embeddedinstance_name, name, value)
...
         end
       end
+
+      def self.reboot_relationship(resource)
+        reboot_resource = resource.catalog.resource(:reboot, 'dsc_reboot')
+        return [] if !reboot_resource
+
+        # if an edge already exists from Reboot[dsc_reboot] to this resource, we have problems
+        if resource.catalog.relationship_graph.edge?(reboot_resource, resource)
+          # TODO: should this be a hard fail? probably?
```
Iristyle: This is my only open question... otherwise, I think that this is ready to go.

Iristyle:
@jpogran / @ferventcoder I tested this on Puppet 3.8.4 and Puppet 4.x in conjunction with the changes to the reboot module in https://github.com/puppetlabs/puppetlabs-reboot/pull/97

I verified that we don't see the warnings like we did previously in #104, and that reboots are triggered only when we set the property on the DSC resource.

Please try your best to break it ;0

Iristyle:
@joshcooper it might be helpful for you take a look at the graph manipulation code as well.  It was interesting finding the right integration point to inject edges (keeping in mind the 3.8 support, and inability to leverage `autonotify`)

lib/puppet/provider/base_dsc/powershell.rb
```diff
@@ -67,8 +71,30 @@ def destroy
...
+    unless reboot_resource
+      Puppet.warning "No reboot resource found in the graph that has 'dsc_reboot' as it's name. Cannot signal reboot to Puppet."
+      return
+    end
+
+    if reboot_resource.provider.respond_to?(:reboot_required)
+      # internal API used to let reboot resource knows a reboot is pending
+      reboot_resource.provider.reboot_required = true
+    else
+      Puppet.warning "Reboot module must be updated, since resource does not have :reboot_required method implemented. Cannot signal reboot to Puppet."
```
jpogran: Should we change the wording for this line? If the reboot resource is present, we do actually reboot the host.

Iristyle: I'm a little :confused: This is the `else` branch, where the reboot resource doesn't have a `reboot_required` member, so a reboot won't be initiated.

lib/puppet/provider/base_dsc/powershell.rb
```diff
@@ -67,8 +71,30 @@ def destroy
...
+
+    reboot_resource = resource.catalog.resource(:reboot, 'dsc_reboot')
+    unless reboot_resource
+      Puppet.warning "No reboot resource found in the graph that has 'dsc_reboot' as it's name. Cannot signal reboot to Puppet."
+      return
+    end
+
+    if reboot_resource.provider.respond_to?(:reboot_required)
+      # internal API used to let reboot resource knows a reboot is pending
+      reboot_resource.provider.reboot_required = true
```
ferventcoder: Hackalicious...

lib/puppet_x/puppetlabs/dsc_type_helpers.rb
```diff
@@ -126,6 +126,24 @@ def self.validate_mof_type(mof_type, embeddedinstance_name, name, value)
           end
         end
       end
+
+      def self.reboot_relationship(resource)
+        reboot_resource = resource.catalog.resource(:reboot, 'dsc_reboot')
+        return [] if !reboot_resource
+
+        # if an edge already exists from Reboot[dsc_reboot] to this resource, we have problems
```
ferventcoder: @Iristyle why is this so?

lib/puppet/provider/base_dsc/powershell.rb
```diff
@@ -67,8 +71,30 @@ def destroy
...
+
+    notify_reboot_pending if data['rebootrequired'] == true
+
+    data
+  end
+
+  def notify_reboot_pending
+    Puppet.info "A reboot is required to progress further. Notifying Puppet."
+
+    reboot_resource = resource.catalog.resource(:reboot, 'dsc_reboot')
```
ferventcoder: When all else fails, result to searching the catalog.

ferventcoder: This must mean we've resorted to the best worst case scenario.

lib/puppet_x/puppetlabs/dsc_type_helpers.rb
ferventcoder: Our instructions for this specifically state a certain way for the user to set this up.

ferventcoder: So a hard failure makes sense for me.

ferventcoder: What happens when multiple resources want to define this as an edge? Because I think that is a totally legit scenario.

lib/puppet/provider/base_dsc/powershell.rb
jpogran: A reboot is initiated regardless

```
Debug: Create Dsc Resource returned: {"rebootrequired":true,"indesiredstate":true,"errormessage":""}

Info: A reboot is required to progress further. Notifying Puppet.
Warning: Reboot module must be updated, since resource does not have :reboot_required method implemented. Cannot signal reboot to Puppet.
Notice: /Stage[main]/Main/Dsc_xcomputer[foo]/ensure: created
Info: /Stage[main]/Main/Dsc_xcomputer[foo]: Scheduling refresh of Reboot[dsc_reboot]
Debug: /Stage[main]/Main/Dsc_xcomputer[foo]: The container Class[Main] will propagate my refresh event
Notice: Scheduling system reboot with message: "Puppet is rebooting the computer"
Debug: Adding C:\Windows\sysnative\shutdown.exe /r /t 60 /d p:4:1 /c "Puppet is rebooting the computer" to ruby's at_exit handler
Notice: /Stage[main]/Main/Reboot[dsc_reboot]: Triggered 'refresh' from 1 events
```

lib/puppet_x/puppetlabs/dsc_type_helpers.rb
ferventcoder: In fact I think when you define with multiple resources that could have edges, this is going to get called.

lib/puppet_x/puppetlabs/dsc_type_helpers.rb
Iristyle: @ferventcoder because it would create a circular dependency (and also doesn't make any sense)

Before we merge this, let's run a test where a user has created a bad manifest like:

```puppet
reboot { 'dsc_reboot':
  when => pending
} ->

dsc_file {'foo':
  dsc_destinationpath => 'c:\\foo.txt',
  dsc_contents => 'foo'
}
```

Iristyle:
@hlindberg it would be great if you could take a look at this as well.

lib/puppet_x/puppetlabs/dsc_type_helpers.rb
Iristyle: This is designed so that multiple DSC resources may direct an edge at the Reboot resource already.  The `TODO` question revolves around whether the user has incorrectly created an edge in the opposite direction.

lib/puppet/provider/base_dsc/powershell.rb
Iristyle: Not sure I understand the comment.  We have 2 independent things happening here:

* When we build up the graph, we draw an edge (by finding the resource)
* When we need to set the `pending` state (which is what this code does), we find the resource, and set a flag


We need the edge in the graph is to ensure we take a look at pending reboot status after each and every DSC resource application.

lib/puppet_x/puppetlabs/dsc_type_helpers.rb
ferventcoder: OH! Right on. I would just remove it ;)

lib/puppet_x/puppetlabs/dsc_type_helpers.rb
ferventcoder: Makes sense now, I think I missed it was going to create a circular dependency, which we most certainly do not want.

lib/puppet/provider/base_dsc/powershell.rb
ferventcoder: I am commenting on the fact that we're now using Puppet internals to get to the catalog resources as a last resort when there are no other options.

lib/puppet_x/puppetlabs/dsc_type_helpers.rb
Iristyle: So the current code allows the run to proceed by not connecting the edge that would create the cyclic dependency.

If we try to create the cyclic dependency, then the run dies with:

```
Error: Failed to apply catalog: Found 1 dependency cycle:
(Dsc_file[foo] => Reboot[dsc_reboot] => Dsc_file[foo])
```

So if we're going to recognize the cycle, we should definitely fail I think.  The question is, do we bother creating a better msg for the user, or just let the standard Puppet error bubble up to them?

Iristyle: I opted to rely on Puppets cyclic dependency checking which is certainly more robust than checking for a single edge, which is what we were doing.

This simplifies the edge building code a little bit as a side effect.

hlindberg:
Was it really necessary to have that may copies of this method? :
```
def builddepends
    super() + PuppetX::Dsc::TypeHelpers.reboot_relationship(self)
end
```
As opposed to metaprogram them in place.

Iristyle:
@hlindberg I'm not too worried about the method duplication given all of these types are generated.

hlindberg:
ok, that makes sense, otherwise it seems like a perfect place where you could get drift. Was there anything else in particular you wanted me to look at @Iristyle ?

Iristyle:
@hlindberg - nope - just wanted a sanity check on the graph manipulation bits in `dsc_type_helpers.rb` / `powershell.rb`, since we had talked a bit about how to handle this sensibly.

The code still feels a bit hacky, but at least it's not a *lot of code* to add edges as necessary, and find the appropriate resource to signal (and it's done defensively)... so if we end up with a better mechanism in the future, shouldn't be too hard to change.

lib/puppet_x/puppetlabs/dsc_type_helpers.rb
```diff
@@ -130,6 +130,23 @@ def self.validate_mof_type(mof_type, embeddedinstance_name, name, value)
...
+      # if an edge already exists from Reboot[dsc_reboot] to this resource
+      # Puppet will recognize the cyclic dependency automatically and fail hard:
+      # Error: Failed to apply catalog: Found 1 dependency cycle:
+      # (Dsc_file[foo] => Reboot[dsc_reboot] => Dsc_file[foo])
+      def self.reboot_relationship(resource)
+        reboot_resource = resource.catalog.resource(:reboot, 'dsc_reboot')
+
+        # do nothing if no Reboot[dsc_reboot] or already an edge from resource to it
+        return [] if (!reboot_resource ||
+          resource.catalog.relationship_graph.edge?(resource, reboot_resource))
```
Iristyle: Thanks to @cowofevil we've discovered this edge check is likely ineffectual / irrelevant at this point in time.  This call should accept the list of other relationships created by `builddepends` and search that instead.  Whoops

Iristyle:
This is ready for another pass @ferventcoder

lib/puppet/provider/base_dsc/powershell.rb
```diff
@@ -67,8 +71,30 @@ def destroy
...
+    unless reboot_resource
+      Puppet.warning "No reboot resource found in the graph that has 'dsc_reboot' as it's name. Cannot signal reboot to Puppet."
```
ferventcoder: Minor nit - `it's` should be `its`

Iristyle: :+1: thanks

lib/puppet_x/puppetlabs/dsc_type_helpers.rb
```diff
@@ -130,6 +130,35 @@ def self.validate_mof_type(mof_type, embeddedinstance_name, name, value)
...
+        # do nothing if no Reboot[dsc_reboot] or already an edge from resource to it
+        if should_add_reboot_relationship(resource, reboot_resource, pending_relationships)
+          # otherwise build resource[name] => Reboot[dsc_reboot]
+          edge = Puppet::Relationship.new(resource, reboot_resource,
+              { :callback => :refresh, :event => :ALL_EVENTS })
+          pending_relationships << edge
+        end
+
+        pending_relationships
+      end
```
ferventcoder: Did I miss specs on this somewhere?

ferventcoder: Disregard, there are acceptance tests that catch this, so I'm good with that. It may be a little hard to test in specs so acceptance makes total sense to catch it.

PR #144, opened 2015-11-17
(WIP)(MODULES-2709) - Host the CLR with COM

Iristyle:
- Starting work for hosting PowerShell

### DIFF
--- a/lib/puppet_x/puppetlabs/clr_host.rb
+++ b/lib/puppet_x/puppetlabs/clr_host.rb
@@ -1,0 +1,513 @@
+require 'puppet/util/windows'
+
+module PuppetX
+  module Dsc
+    class CLRHost
+      include Puppet::Util::Windows::String
+
+      require 'ffi'
+      extend FFI::Library
+
+      # The error class raised if any CLR calls fail.
+      class Error < Puppet::Util::Windows::Error; end
+
+      private
+
+      # TODO: this line didn't appear to take affect.. hmmm
+      # com = Puppet::Util::Windows::COM
+
+      class << self
+        attr_accessor :com_initialized
+      end
+
+      # NOTE: a bunch of this stuff comes from the Windows SDK
+      # C:\Program Files (x86)\Microsoft SDKs\Windows\v7.1A\Include\mscoree.h
+      # C:\Program Files (x86)\Microsoft SDKs\Windows\v7.1A\Include\metahost.h
+      # C:\Program Files (x86)\Microsoft SDKs\Windows\v7.1A\Include\ObjIdl.h
+
+      # WIN32 API
+      # https://msdn.microsoft.com/en-us/library/dd537633(v=vs.110).aspx
+      # HRESULT CLRCreateInstance(
+      #     [in]  REFCLSID  clsid,
+      #     [in]  REFIID     riid,
+      #     [out] LPVOID  * ppInterface
+      # );
+      ffi_lib :mscoree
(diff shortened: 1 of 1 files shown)


Iristyle:
@jpogran / @ferventcoder still have some work to do here... but not too far from getting CLR completely loaded through COM

ferventcoder:
Neat.

lib/puppet_x/puppetlabs/clr_host.rb
```diff
@@ -0,0 +1,526 @@
...
+      # # HRESULT error codes
+      # # http://blogs.msdn.com/b/eldar/archive/2007/04/03/a-lot-of-hresult-codes.aspx
+      # # in Ruby, an 0x8XXXXXXX style HRESULT can be resolved to 2s complement
+      # # by using "0x8XXXXXXX".to_i(16) - - 0x100000000
+      # SCHED_E_ACCOUNT_INFORMATION_NOT_SET   = -2147216625 # 0x8004130F
+      # SCHED_E_NO_SECURITY_SERVICES          = -2147216622 # 0x80041312
+      # # No mapping between account names and security IDs was done.
+      # ERROR_NONE_MAPPED                     = -2147023564 # 0x80070534  WIN32 Error CODE 1332 (0x534)
+
+      public
```
ferventcoder: First time I've seen public after private.

Iristyle: Really? ;0

https://github.com/puppetlabs/puppet/blob/master/lib/puppet/util/windows/taskscheduler.rb#L15-L95

lib/puppet_x/puppetlabs/clr_host.rb
```diff
@@ -0,0 +1,510 @@
...
+        @pClrRuntimeHost.Start()
+
+        @pClrRuntimeHost
+      end
+
+      # TODO: demos used a custom C# class CSSimpleObject.GetStringLength
+      # TODO: C# signature must be:
+      # TODO: static int pwzMethodName (String pwzArgument)
+      # TODO: but if we do that, we have to compile one on the fly somehow!
+      def execute_method(assembly_path, type_name, static_method_name, argument)
```
Iristyle: @ferventcoder / @jpogran I think this ends up being a dead end.  The method signature is too restrictive, unless we have another way to communicate with the loaded runtime.

No way to pass JSON back without some sort of IPC.

lib/puppet_x/puppetlabs/clr_host.rb
ferventcoder: First time "I've" seen it ;)

Iristyle: Fo really really? ;0

https://github.com/puppetlabs/puppet/pull/2848

Iristyle:
Abandon ship (for now anyhow)... this isn't going to work without being able to:

- Ship an assembly OR compile a .NET asm on the fly (not *too hard* with csc.exe), but this poses some security issues
- Establish some sort of lightweight RPC b/w the hosted CLR process and Ruby (keeping in mind, the .NET entry point essentially news up a worker thread to accept requests and returns immediately)
- The best communication method is probably a named pipe for security reasons - so the idea would be to return the handle from [CreateNamedPipe](https://msdn.microsoft.com/en-us/library/windows/desktop/aa365150\(v=vs.85\).aspx) as the int, so that host Ruby can establish bi-directional comms

Not impossible, but a little tricky / out of scope for now

PR #147, opened 2015-11-17
(MODULES-2520) Validate readonly params by default

ferventcoder:
If the value of the readonly parameter is not what we expect it
to be, fail validation for the parameter. Unfortunately using
properties and overriding property.retrieve did not work as intended.

Fallback to a method that does work appropriately. Validate the readonly
parameter is the same as the default value.


build/dsc/templates/dsc_type.rb.erb
```diff
@@ -74,28 +74,24 @@ Puppet::Type.newtype(:dsc_<%= resource.friendlyname.downcase %>) do
...
+  newparam(:dscmeta_resource_friendly_name) do
     desc "A read-only value that is the DSC Resource Friendly Name ('<%= resource.friendlyname %>')."
 
-    def retrieve
-      '<%= resource.friendlyname %>'
-    end
-
     validate do |value|
-      fail 'dscmeta_resource_friendly_name is read-only'
+      fail 'dscmeta_resource_friendly_name is read-only' if value != '<%= resource.friendlyname %>'
```
Iristyle: I think you can remove the `if` ... lets make it simpler and just fail if a user tries to set it at all (which is the method we use in AWS module - i.e. https://github.com/puppetlabs/puppetlabs-aws/blob/master/lib/puppet/type/ec2_instance.rb)

ferventcoder: That is exactly what we were trying to do before.

ferventcoder: The problem is defaultto also counts as the user setting it.

Iristyle: Gotcha, going to try to some things out...

Iristyle: OK - so I think we can actually simplify this whole thing quite a bit, but departing from prior art.  Let me PR your PR to show you what I mean.

Iristyle:
Superceded by #149

PR #136, opened 2015-11-06
(MODULES-2777) Build all DSC Types

jpogran:
Update all types to the latest import from the DSC Resources repo to sha 84a467c

PR #149, opened 2015-11-18
(MODULES-2520) Fix type templates / regenerate

Iristyle:
- In #89 some changes
   were made around type generation for the sake of documentation.

   Specifically, 8b1415e changed some
   Puppet parameters to properties in an effort to make them
   read-only.  In retrospect, there were 2 issues with the merge of
   this PR:

   - The types were not re-generated and run through acceptance tests
     to verify nothing broke as a result of these changes.  When a later
     PR #136 was merged
     in, commit b689baf dragged in
     these latent changes.  This commit was reverted.
   - These sorts of functional changes should have been isolated from
     the documentation changes proposed in the original PR.

- As it turned out, the change from parameters to properties completely
   broke the PowerShell template generation being used with
   Invoke-DscResource (hence the revert)

- This commit changes the properties to actual members of the type
   since there isn't any compelling reason to surface them as a Puppet
   property / parameter anyhow.  They are strictly internal
   implementation details, so don't need additional Puppet overhead or
   visibility.  The Invoke-DscResource template has been updated
   accordingly to access these values differently.

- All types have been regenerated to incorporate in both the previous
   codegen changes that were never introduced in the prior PR, and the
   new changes for dscmeta_resource_friendly_name, dscmeta_resource_name,
   dscmeta_module_name and dscmeta_module_version


Iristyle:
Ok @ferventcoder - test results

```
      Test Suite: tests @ 2015-11-17 18:19:56 -0800

       Total Suite Time: 2105.54 seconds
      Average Test Time: 31.43 seconds
              Attempted: 67
                 Passed: 67
                 Failed: 0
                Errored: 0
                Skipped: 0
                Pending: 0
                  Total: 67

      - Specific Test Case Status -
```

PR #148, opened 2015-11-17
(MODULES-2799) Build custom types

ferventcoder:
Updating the build to support building custom types
again.

PR #151, opened 2015-11-20
(MODULES-2837) Specify DSC Resource Module Path

jpogran:
We ship the DSC Resources in the box that we build our types on. We used
to have to symlink
$env:ProgramFiles\WindowsPowerShell\Modules\PuppetVendoredModules to the
install path of our module in order for DSC to 'find' our vendored DSC
Resources Correctly.

With these changes we can now have our DSC Resources stored inside our
module and point Invoke-DscResource to them. This also solves the
duplicate DSC Resource problem of having the same DSC Resources in the
Modules folder.

We accomplish this by resolving the exact path to the manifest file for
the DSC Resource module and passing that to Invoke-DscResource.

- [X] Resolve path to PS manifest file
- [X] Add manifest to PS invoke script
- [X] Remove symlink functionality
- [x] Test with custom built DSC Resource  #148 
- [x] Fix up acceptance to remove any validation surrounding symlink creation (if there are any)


lib/puppet/provider/templates/invoke_dsc_resource.ps1.erb
```diff
@@ -51,7 +51,7 @@ $invokeParams = @{
 <% end -%>
   }<% if resource.respond_to?(:dscmeta_module_version) %>
   ModuleName = @{
-    ModuleName      = <%= "\"#{resource.dscmeta_module_name}\"" %>
+    ModuleName      = <%= "\"#{vendored_modules_path}\\#{resource.dscmeta_module_name}\\#{resource.dscmeta_module_name}.psd1\"" %>
```
ferventcoder: Seems legit.

lib/puppet/provider/base_dsc/powershell.rb
```diff
@@ -20,6 +20,10 @@
 Applies DSC Resources by generating a configuration file and applying it.
 EOT
 
+  def vendored_modules_path
+    File.expand_path(Pathname.new(__FILE__).dirname + '../../../' + 'puppet_x/dsc_resources').gsub(/\//,'\\')
```
Iristyle: I don't believe that switching the slashes is necessary, is it?

Iristyle:
Huge :+1: to this ... looking forward to merging it.

/cc @cowofevil

lib/puppet/provider/base_dsc/powershell.rb
jpogran: It was necessary because i put backslashes in the erb template. Replacing those with forward slashes removes the need for it. Fixing now.

lib/puppet/feature/vendors_dsc.rb
```diff
@@ -2,10 +2,11 @@
...
-    true
-  else
-    false
-  end
+  # if Puppet::Util::Platform.windows?
+  #   DscSymlink.set_symlink_folder
+  #   true
+  # else
+  #   false
+  # end
```
jpogran: How should I disable the symlink? Removing this file requires removing the entire test suite. Returning false works but seems silly.

Iristyle: IMHO, just axe all symlink related stuff, including tests + acceptance tests.

ferventcoder: Yep, that's what I was thinking as well.

ferventcoder: Any acceptance tests that need to also be fixed up by this?

lib/puppet/provider/templates/invoke_dsc_resource.ps1.erb
```diff
@@ -51,7 +51,7 @@ $invokeParams = @{
 <% end -%>
   }<% if resource.respond_to?(:dscmeta_module_version) %>
   ModuleName = @{
-    ModuleName      = <%= "\"#{resource.dscmeta_module_name}\"" %>
+    ModuleName      = <%= "\"#{vendored_modules_path}/#{resource.dscmeta_module_name}/#{resource.dscmeta_module_name}.psd1\"" %>
```
ferventcoder: We need to check if it is a built-in one that we are not vendoring and just use the name instead.

ferventcoder: I'm thinking we might want to be just a bit more explicit on this check though.

jpogran:
Acceptance tests pass:

```
     Test Suite: tests @ 2015-11-23 12:10:15 -0500

       Total Suite Time: 2496.51 seconds
      Average Test Time: 37.26 seconds
              Attempted: 67
                 Passed: 67
                 Failed: 0
                Errored: 0
                Skipped: 0
                Pending: 0
                  Total: 67

      - Specific Test Case Status -

Failed Tests Cases:
Errored Tests Cases:
Skipped Tests Cases:
Pending Tests Cases:
```

lib/puppet/provider/templates/invoke_dsc_resource.ps1.erb
Iristyle: Not sure I understand, since we check for existence of `dscmeta_module_version` (which is only generated on types that are not a part of `PSDesiredStateConfiguration` per https://github.com/puppetlabs/puppetlabs-dsc/blob/master/build/dsc/templates/dsc_type.rb.erb#L80-L82

ferventcoder: Right, but looking at this it's not explicit about what the branching is done for. We know that it is for `dscmeta_module_version` but that means what when you are looking at this out of context? :)

PR #152, opened 2015-11-20
(MODULES-2853) Update Broken Integration Tests

cowofevil:
The duplicate module integration tests were failing on PE 3.8.x because of
an erroneous assumption about paths for the Puppet module cache.
acceptance test
[skip ci]

### DIFF
--- a/tests/integration/tests/user_scenarios/duplicate_different_module_version.rb
+++ b/tests/integration/tests/user_scenarios/duplicate_different_module_version.rb
@@ -6,11 +6,16 @@ test_name 'MODULES-2743 - C94804 - Apply DSC Manifest that Uses a Resource from 
 # Init
 local_files_root_path = ENV['MANIFESTS'] || 'tests/manifests'
 
-dsc_source_module_path = 'C:\ProgramData\PuppetLabs\puppet\cache\lib\puppet_x\dsc_resources\xPSDesiredStateConfiguration'
+puppet_agent_cache_path = 'C:\ProgramData\PuppetLabs\puppet\cache\lib\puppet_x'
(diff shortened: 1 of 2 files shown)


tests/integration/tests/user_scenarios/duplicate_different_module_version.rb
```diff
@@ -6,12 +6,17 @@
 # Init
 local_files_root_path = ENV['MANIFESTS'] || 'tests/manifests'
 
-dsc_source_module_path = 'C:\ProgramData\PuppetLabs\puppet\cache\lib\puppet_x\dsc_resources\xPSDesiredStateConfiguration'
+puppet_agent_cache_path = 'C:\ProgramData\PuppetLabs\puppet\cache\lib\puppet_x'
```
Iristyle: We should be asking Puppet for these values instead of hardcoding them - i.e.  `agent.puppet['vardir']`

tests/integration/tests/user_scenarios/duplicate_same_module_version.rb
```diff
@@ -6,10 +6,15 @@
 # Init
 local_files_root_path = ENV['MANIFESTS'] || 'tests/manifests'
 
-dsc_source_module_path = 'C:\ProgramData\PuppetLabs\puppet\cache\lib\puppet_x\dsc_resources\xPSDesiredStateConfiguration'
+puppet_agent_cache_path = 'C:\ProgramData\PuppetLabs\puppet\cache\lib\puppet_x'
```
Iristyle: Same thing with paths here - Beaker has helpers for this kind of stuff already based on it's defaults.  And you can always ask the agent for the value.

PR #153, opened 2015-11-23
(maint) Pin mime-types version to allow Ruby 1.9

ferventcoder:
puppet-blacksmith depends on mime-types gem, which just released
a new version that requires Ruby 2. We require building the module
with 1.9.3 so pin the version to 2.99 which will allow using Ruby
less than 2.0.

### DIFF
--- a/.sync.yml
+++ b/.sync.yml
@@ -17,6 +17,8 @@ Gemfile:
       - gem: mocha
         version: '~>0.10.5'
       - gem: puppet-blacksmith
+      - gem: mime-types
+        version: '~>2.99'
     ':system_tests':
       - gem: beaker
       - gem: master_manipulator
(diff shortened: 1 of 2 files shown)


.sync.yml
```diff
@@ -17,6 +17,8 @@ Gemfile:
       - gem: mocha
         version: '~>0.10.5'
       - gem: puppet-blacksmith
+      - gem: mime-types
+        version: '~> 2.99'
```
Iristyle: Minor nit, but I think you want to skip the space like `'~>2.99'` to match what you've put in the `Gemfile`

ferventcoder: sure can

ferventcoder: fixed.

PR #155, opened 2015-12-04
(maint) Move PowerShell provider methods to class

Iristyle:
- Previously, PowerShell provider methods were all instance methods,
   but this isn't strictly necessary.  Move methods for generating the
   Invoke-DscResource script to the providers class, so that they may
   later be called by self.instances / self.prefetch when adding 'get'
   support.


Iristyle:
@ferventcoder this is a precursor to supporting `get` over DSC resources.  I figured we could get merging this out of the way now, since it's a pretty easy change to wrap your head around.

ferventcoder:
LGTM

PR #156, opened 2015-12-04
(MODULES-2843) Create Tests for Reboot

cowofevil:
Add integration tests for the reboot feature of the DSC module. This entailed
adding a special DSC resource "PuppetFakeResource" for testing. Also, the
"dsc_utils.rb" library was updated to add new helper methods for installing
and uninstalling said special DSC resource.

acceptance test
[skip ci]


Iristyle:
Please save the files `PuppetFakeResource.psm1`, `PuppetFakeResource.schema.mof` and `PuppetFakeResource.psd1` as UTF-8 with no BOM and update the PR.  I believe the files are the PS default `UCS-2`

ferventcoder:
@Iristyle Looks like this has been done. :)

PR #163, opened 2015-12-21
(MODULES-2512) Cannot ensure absent on Service DSC Resource

jpogran:
If a user specifies a service as absent, when it is not running, it will
not be removed. This is because the dsc resource returns false when testing for
existence instead of true even though service exists because of the
following code `return ($State -eq "Stopped" -and $svc.Status -eq
"Stopped") -or ($svc.Status -eq "Running" -and $State -eq "Running")`
Since $state is a paramter and is defautled to 'Running', if the service
is stopped this will always return false instead of true.

If a user specifies a service as absent when the service is already not
present, it will error. This happens because we invert absent to present
to determine if resource exists. t\This causes the error because testing
for present requires a Path, but not when testing it is Absent. Since we
dont provide the path there is nothing to pass to Present, so it throws
an error during the testing phase.

This updates the readme to inform the user of workarounds to these
issues.

### DIFF
--- a/README.md
+++ b/README.md
@@ -334,7 +334,39 @@ information, see about_Execution_Policies at http://go.microsoft.com/fwlink/?Lin
 
 - The `dsc_log` resource may not appear to work. The ["Log" resource](https://technet.microsoft.com/en-us/library/Dn282117.aspx) writes events to the 'Microsoft-Windows-Desired State Configuration/Analytic' event log, which is [disabled by default](https://technet.microsoft.com/en-us/library/Cc749492.aspx).
 
-- You may have issues attempting to use `dsc_ensure => absent` on `dsc_service`. See [MODULES-2512](https://tickets.puppetlabs.com/browse/MODULES-2512) for details.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -334,7 +334,39 @@ information, see about_Execution_Policies at http://go.microsoft.com/fwlink/?Lin
...
+When setting resources to absent, it is common to specify a minimal statement like so:
+
+~~~
+dsc_service{'disable_foo':
+  dsc_ensure => absent,
+  dsc_name => 'foo'
+}
+~~~
+
+Due to the way the Service DSC Resource sets it's defaults, the above minimal statement will report the service as already absent erroneously if the service is currently not running. In order to work around this, you must specify that `State => 'Stopped'` as well as `Ensure => absent'`. The following example will work:
```
ferventcoder: its versus it's

ferventcoder:
Other than the grammar bit, it LGTM.

jpogran:
Grammar fixed

PR #164, opened 2015-12-21
(MODULES-2919) DSC - Intermittent failure of Windows Process Tests

glennsarti:
Modified the command used to during the test and increased the execution time
from as-fast-as-possible to just over 1 second.  This will help stop false CI
failures caused by test hosts under heavy load


ferventcoder:
LGTM

PR #165, opened 2015-12-30
(MODULES-2709) Reuse PowerShell instance for DSC

Iristyle:
- Currently, each time a resource modification is made, the PowerShell
   provider opens up new powershell.exe instance(s), makes a single
   Invoke-DscResource call, then terminates.

   This is a particularly expensive endeavor, since each resource
   requires a minimum of one powershell.exe instance for the sake of
   using Test to examine if the DSC resource is in the desired state. If
   a resource needs to be added / removed / modified via the Set method
   of Invoke-DscResource, then an additional instance is required.

   With the working set and slow startup times of powershell.exe this
   is a particularly expensive cost to pay per run.

 - Add a new PowerShellManager class that supports reuse of an initially
   created powershell.exe instance.  Modify the PowerShell provider to
   use this helper class instead of running powershell.exe on demand.

   In simple tests, this significantly (by a factor or 2x or more)
   reduces the runtime of a manifest.

 - The PowerShellManager class uses popen2 since the PowerShell template
   currently handles capturing Invoke-DscResource errors and propagating
   them through the JSON that is returned to Ruby.  Therefore, ignore
   stderr stream.  In other exploratory work, there was difficulty in
   getting stderr to be read from properly - in such instances, the
   IO.select call over stderr determined that the stream was readable,
   but the subsequent call to stderr.gets would block, and deadlock. Its
   presumed that this is a problem specifically with Ruby on Windows, or
   with capturing stderr from powershell.exe - but the exact cause has
   not yet been determined.

### DIFF
--- a/lib/puppet/provider/base_dsc/powershell.rb
+++ b/lib/puppet/provider/base_dsc/powershell.rb
@@ -21,6 +21,25 @@ Puppet::Type.type(:base_dsc).provide(:powershell) do
 Applies DSC Resources by generating a configuration file and applying it.
 EOT
 
+  def self.upgrade_message
+    Puppet.warning <<-UPGRADE
+The current Puppet version is outdated and uses a library that was
+previously necessary on the current Ruby verison to support a colored console.
+
+Unfortunately this library prevents the DSC module from using a shared
+PowerShell process to dramatically improve the performance of resource
+application.
+
+To enable these improvements, it is suggested to upgrade to any x64 version of
+Puppet (including 3.x), or to a Puppet version newer than 3.x.
+    UPGRADE
+  end
+
+
+  if Facter.value(:uses_win32console)
+    upgrade_message
+  end
+
   def self.vendored_modules_path
     File.expand_path(Pathname.new(__FILE__).dirname + '../../../' + 'puppet_x/dsc_resources')
   end
@@ -35,9 +54,10 @@ EOT
     File.expand_path('../../templates', __FILE__)
   end
 
-
   def self.powershell_args
-    ['-NoProfile', '-NonInteractive', '-Sta', '-NoLogo', '-ExecutionPolicy', 'Bypass', '-Command', '-']
+    ps_args = ['-NoProfile', '-NonInteractive', '-Sta', '-NoLogo', '-ExecutionPolicy', 'Bypass', '-Command']
+    ps_args << '-' if !Facter.value(:uses_win32console)
+    ps_args
   end
 
   def ps_manager
@@ -46,10 +66,15 @@ EOT
 
   def exists?
     version = Facter.value(:powershell_version)
+    uses_win32console = Facter.value(:uses_win32console)
     Puppet.debug "PowerShell Version: #{version}"
     script_content = ps_script_content('test')
     Puppet.debug "\n" + script_content
-    output = ps_manager.execute(script_content)[:stdout]
+    if uses_win32console
+      output = powershell(self.class.powershell_args, script_content)
+    else
+      output = ps_manager.execute(script_content)[:stdout]
+    end
     Puppet.debug "Dsc Resource returned: #{output}"
     data = JSON.parse(output)
     fail(data['errormessage']) if !data['errormessage'].empty?
@@ -61,9 +86,14 @@ EOT
   end
 
   def create
+    uses_win32console = Facter.value(:uses_win32console)
     script_content = ps_script_content('set')
     Puppet.debug "\n" + script_content
-    output = ps_manager.execute(script_content)[:stdout]
+    if uses_win32console
+      output = powershell(self.class.powershell_args, script_content)
+    else
+      output = ps_manager.execute(script_content)[:stdout]
+    end
     Puppet.debug "Create Dsc Resource returned: #{output}"
     data = JSON.parse(output)
 
@@ -75,9 +105,14 @@ EOT
   end
 
   def destroy
+    uses_win32console = Facter.value(:uses_win32console)
     script_content = ps_script_content('set')
     Puppet.debug "\n" + script_content
-    output = ps_manager.execute(script_content)[:stdout]
+    if uses_win32console
+      output = powershell(self.class.powershell_args, script_content)
+    else
+      output = ps_manager.execute(script_content)[:stdout]
+    end
     Puppet.debug "Destroy Dsc Resource returned: #{output}"
     data = JSON.parse(output)
 
--- a/lib/puppet/provider/base_dsc/powershell.rb
+++ b/lib/puppet/provider/base_dsc/powershell.rb
@@ -1,5 +1,8 @@
 require 'pathname'
 require 'json'
+if Puppet::Util::Platform.windows?
+  require_relative '../../../puppet_x/puppetlabs/powershell_manager'
+end
 
 Puppet::Type.type(:base_dsc).provide(:powershell) do
   confine :operatingsystem => :windows
@@ -34,7 +37,11 @@ EOT
 
 
   def self.powershell_args
-    ['-NoProfile', '-NonInteractive', '-NoLogo', '-ExecutionPolicy', 'Bypass', '-Command']
+    ['-NoProfile', '-NonInteractive', '-Sta', '-NoLogo', '-ExecutionPolicy', 'Bypass', '-Command', '-']
+  end
+
+  def ps_manager
+    PuppetX::Dsc::PowerShellManager.instance("#{command(:powershell)} #{self.class.powershell_args.join(' ')}")
   end
 
   def exists?
@@ -42,7 +49,7 @@ EOT
     Puppet.debug "PowerShell Version: #{version}"
     script_content = ps_script_content('test')
     Puppet.debug "\n" + script_content
-    output = powershell(self.class.powershell_args, script_content)
+    output = ps_manager.execute(script_content)[:stdout]
     Puppet.debug "Dsc Resource returned: #{output}"
     data = JSON.parse(output)
     fail(data['errormessage']) if !data['errormessage'].empty?
@@ -56,7 +63,7 @@ EOT
   def create
     script_content = ps_script_content('set')
     Puppet.debug "\n" + script_content
-    output = powershell(self.class.powershell_args, script_content)
+    output = ps_manager.execute(script_content)[:stdout]
     Puppet.debug "Create Dsc Resource returned: #{output}"
     data = JSON.parse(output)
 
@@ -70,7 +77,7 @@ EOT
   def destroy
     script_content = ps_script_content('set')
     Puppet.debug "\n" + script_content
-    output = powershell(self.class.powershell_args, script_content)
+    output = ps_manager.execute(script_content)[:stdout]
     Puppet.debug "Destroy Dsc Resource returned: #{output}"
     data = JSON.parse(output)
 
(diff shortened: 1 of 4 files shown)


Iristyle:
@puppetlabs/windows needs some tests and a few more tweaks, but have at it!

ferventcoder:
Weird, this didn't show up in notifications.

lib/puppet_x/puppetlabs/powershell_manager.rb
```diff
@@ -0,0 +1,123 @@
...
+    end
+
+    def execute(powershell_code)
+      # always need a trailing newline to ensure PowerShell parses code
+      out = exec_read_result(<<-CODE
+      $Error.Clear()
+      $LASTEXITCODE = 0
+      #{powershell_code}
+      [Console]::Out.Flush()
+      [Console]::Error.Flush()
```
ferventcoder: And we don't need to grab the last result because we are just looking at the output, correct?

Iristyle: Yeah, in a general purpose solution we would want to do that... but here, it's not necessary.

In fact, we can probably ditch `$Error.Clear()` and `$LASTEXITCODE = 0` as well.  I tried making the code submit an `echo $LASTEXITCODE` call after running the incoming `powershell_code` - but that hung again.

Something really funky is at play with the threads and calling `IO.select`

lib/puppet_x/puppetlabs/powershell_manager.rb
```diff
@@ -0,0 +1,123 @@
...
+            proc_thread.join
+            stdout_monitor_thread.join
+            @return_code = proc_thread.value
+          end
+        end
+      end
+
+      # block initializer until process started with popen
+      process_init_lock.synchronize do
+        process_started.wait(process_init_lock)
```
ferventcoder: something about this feels wrong?

ferventcoder: Wait, chalk it up to confusing myself. haha

lib/puppet_x/puppetlabs/powershell_manager.rb
```diff
@@ -0,0 +1,123 @@
...
+
+      process_init_lock.synchronize do
+        @runner_thread = Thread.new do
+          Open3::popen2(cmd) do |stdin, stdout, proc_thread|
+            Puppet.debug "#{Time.now} #{cmd} is running as pid: #{proc_thread.pid}"
+
+            @stdin = stdin
+            @stdout = stdout
+            stdin.sync = true
+            stdout.sync = true
```
ferventcoder: During rspec runs, the process has started and already exited by the time it hits this line of code, so it hangs once it gets down to waiting on the process to finish opening.

ferventcoder: We probably need to do some immediate error handling if the process exits early, otherwise it will hang forever waiting on the process to start up.

ferventcoder: @jpogran had a malformed command at https://github.com/puppetlabs/puppetlabs-dsc/pull/166/files#r48672603, so the process was exiting as soon as it opened. Which causes the specs to hang while it waited the process to start.

Iristyle:
@puppetlabs/windows this needs some tests, etc... but works now after changing from `mutex` to `event`

ferventcoder:
Sweet.

lib/puppet_x/puppetlabs/powershell_manager.rb
```diff
@@ -0,0 +1,167 @@
...
+          name_ptr)
+
+        if handle == FFI::Pointer::NULL_HANDLE
+          msg = "Failed to create new event #{name}"
+          raise Puppet::Util::Windows::Error.new(msg)
+        end
+      end
+
+      handle
+    end
```
ferventcoder: So events instead of mutexes... if it works then :+1:

lib/puppet_x/puppetlabs/powershell_manager.rb
```diff
@@ -0,0 +1,169 @@
...
+      def self.instance(cmd)
+        @@instances[:cmd] ||= PowerShellManager.new(cmd)
+      end
+
+      def initialize(cmd)
+        @output_ready_event_name =  "Global\\#{SecureRandom.uuid}"
+        @output_ready_event = self.class.create_event(@output_ready_event_name)
+
+        @stdin, @stdout, @ps_process = Open3.popen2(cmd)
+        @stdin.sync = true
```
Iristyle: `.sync` should be unnecessary

lib/puppet_x/puppetlabs/powershell_manager.rb
```diff
@@ -0,0 +1,169 @@
...
+
+        at_exit { exit }
+      end
+
+      def execute(powershell_code)
+        # always need a trailing newline to ensure PowerShell parses code
+        out = exec_read_result(<<-CODE
+        $event = [Threading.EventWaitHandle]::OpenExisting("#{@output_ready_event_name}")
+
+        $Error.Clear()
```
Iristyle: We're not using `$Error` or `$LASTEXITCODE`, so clearing them is not necessary

lib/puppet_x/puppetlabs/powershell_manager.rb
```diff
@@ -0,0 +1,169 @@
...
+        $LASTEXITCODE = 0
+
+        #{powershell_code}
+
+        [Void]$event.Set()
```
Iristyle: If `#{powershell_code}` fails to parse or execute for some reason, it's possible the event never gets set here.  If it never gets set, Puppet will deadlock.

We should carefully review the parameters added to the Invoke-DscResource PowerShell template to ensure that there's no opportunity for injection here as well.

lib/puppet_x/puppetlabs/powershell_manager.rb
```diff
@@ -0,0 +1,169 @@
...
+        @stdin, @stdout, @ps_process = Open3.popen2(cmd)
+        @stdin.sync = true
+        @stdout.sync = true
+
+        Puppet.debug "#{Time.now} #{cmd} is running as pid: #{@ps_process[:pid]}"
+
+        at_exit { exit }
+      end
+
+      def execute(powershell_code)
```
Iristyle: Surface `timeout` here with a default, so that as a last resort we can surface this to end users.

lib/puppet_x/puppetlabs/powershell_manager.rb
ferventcoder: Agreed. Unless we think there may be a lingering effect from a previous call.

lib/puppet_x/puppetlabs/powershell_manager.rb
ferventcoder: I'm not sure wrapping the statements in a try/catch/finally will work here if the parser fails, although it may have been fixed in newer versions of POSH (v4+).

ferventcoder: @jpogran do you know the answer to this one?

jpogran: We already wrap `Invoke-DscResource` in a try/catch in these lines https://github.com/puppetlabs/puppetlabs-dsc/blob/master/lib/puppet/provider/templates/invoke_dsc_resource.ps1.erb#L62-L67, so a try/catch block works.

This however is after parameter assignment and after we create a credential, both operations which may throw exceptions.

I'm generally hesitant on adding general catch alls, but it may be worth removing the try/catch from inside the invoke script and moving it out to here, so we can use finally to ensure the event is set.

lib/puppet/provider/base_dsc/powershell.rb
```diff
@@ -34,15 +37,41 @@ def self.template_path
...
 
   def exists?
     version = Facter.value(:powershell_version)
+    uses_win32console = Facter.value(:uses_win32console)
     Puppet.debug "PowerShell Version: #{version}"
     script_content = ps_script_content('test')
     Puppet.debug "\n" + script_content
-    output = powershell(self.class.powershell_args, script_content)
+    if uses_win32console
+      Puppet.warning self.class.upgrade_message
```
ferventcoder: No way to turn off the warning? This is going to affect all users of Puppet 3.x on x86 - it may start to get somewhat annoying to see the message TWICE for every DSC resource.

ferventcoder: Other than that, I'm :+1:

Iristyle: Yeah, agree that the warning is probably best to emit once per run...

Will change after lunch.

Iristyle:
@puppetlabs/windows all :green_heart: !

ferventcoder:
Sweet - will be all over this momentarily

lib/puppet/provider/base_dsc/powershell.rb
```diff
@@ -18,6 +21,25 @@
...
+
+To enable these improvements, it is suggested to upgrade to any x64 version of
+Puppet (including 3.x), or to a Puppet version newer than 3.x.
+    UPGRADE
+  end
+
+
+  if Facter.value(:uses_win32console)
+    upgrade_message
+  end
```
ferventcoder: Much better.

lib/puppet_x/puppetlabs/powershell_manager.rb
```diff
@@ -0,0 +1,197 @@
...
+        # once signaled, ensure everything has been drained
+        output << drain_stdout
+
+        output.join('')
+      rescue => e
+        msg = "Error reading STDOUT: #{e}"
+        raise Puppet::Util::Windows::Error.new(msg)
+      end
+
+      def exec_read_result(powershell_code, output_ready_event, timeout_ms = 120 * 1000)
```
ferventcoder: As I mentioned before I'm not sure this is a high enough amount of time - I'd suggest we make it configurable somehow?

ferventcoder: That can come afterwards though.

Iristyle: Yeah and I agreed @ferventcoder - but didn't update when I pulled in @jpogran commit.  What do you think we should set the max to?  `300` seconds?

ferventcoder: Is there anything that would take more than 5 minutes? I am thinking more like 10 or longer. Choco does a default timeout on commands at a ludicrous default of 45 minutes. That way if there is a download on a slow connection or if the install takes forever, it has a chance to finish.

I'm thinking something like 5 -10 minutes?

Iristyle:
```
      Test Suite: tests @ 2016-01-12 23:42:59 -0800

       Total Suite Time: 1964.37 seconds
      Average Test Time: 29.32 seconds
              Attempted: 67
                 Passed: 67
                 Failed: 0
                Errored: 0
                Skipped: 0
                Pending: 0
                  Total: 67
```

ferventcoder:
Test Suite: tests @ 2016-01-13 01:49:17 -0600

       Total Suite Time: 2019.87 seconds
      Average Test Time: 30.15 seconds
              Attempted: 67
                 Passed: 67
                 Failed: 0
                Errored: 0
                Skipped: 0
                Pending: 0
                  Total: 67

      - Specific Test Case Status -

Failed Tests Cases:
Errored Tests Cases:
Skipped Tests Cases:
Pending Tests Cases:

PR #168, opened 2016-01-07
(MODULES-2709) Single PowerShell Session Tests

jpogran:


lib/puppet_x/puppetlabs/powershell_manager.rb
```diff
@@ -0,0 +1,169 @@
...
+      def self.instance(cmd)
+        @@instances[:cmd] ||= PowerShellManager.new(cmd)
+      end
+
+      def initialize(cmd)
+        @output_ready_event_name =  "Global\\#{SecureRandom.uuid}"
+        @output_ready_event = self.class.create_event(@output_ready_event_name)
+
+        @stdin, @stdout, @ps_process = Open3.popen2(cmd)
+        @stdin.sync = true
```
joshcooper: ruby already sets sync for the stdin pipe that ruby writes to: https://github.com/ruby/ruby/blob/trunk/lib/open3.rb#L142

lib/puppet_x/puppetlabs/powershell_manager.rb
```diff
@@ -0,0 +1,169 @@
...
+        @stdout.sync = true
```
joshcooper: I don't think this is necessary, because ruby only reads from this pipe.

jpogran:
appveyor failure was due to build timeout.

lib/puppet/provider/base_dsc/powershell.rb
```diff
@@ -1,5 +1,6 @@
 require 'pathname'
 require 'json'
+require_relative '../../../puppet_x/puppetlabs/powershell_manager'
```
joshcooper: I think require_relative is only available in ruby 1.9, does the DSC module require a specific puppet or ruby version so this doesn't blow up if someone was to (unlikely) run on puppet < 3.2.0

jpogran: The metadata.json requires >= 3.5.0 for puppet and >= 3.3.0 for PE so I think we prevent that. Is there somewhere else to check?

lib/puppet_x/puppetlabs/powershell_manager.rb
jpogran: Ethan commented the same. Will address shortly.

lib/puppet_x/puppetlabs/powershell_manager.rb
jpogran: Ethan commented the same. Will address shortly.

lib/puppet_x/puppetlabs/powershell_manager.rb
```diff
@@ -0,0 +1,177 @@
...
+        msg = "Error writing STDIN / reading STDOUT: #{e}"
+        raise Puppet::Util::Windows::Error.new(msg)
+      end
+
+      def read_stdout
+        output = []
+        self.class.wait_on(@output_ready_event)
+
+        initially_readable = false
+        while self.class.is_readable?(@stdout, 0.1) do
```
joshcooper: This seems racy in that the powershell process may fill the named pipe buffer, the ruby process then sees the pipe is readable, and drains the pipe (in a single `gets` call below). Ruby ruby loops back around to the `is_readable?` call, and may timeout before the powershell process is scheduled to run and writes more data to the pipe. It may be better to have the powershell side signal an `output_done_event` and have the ruby side keep draining the pipe until the event is signaled.

lib/puppet/provider/base_dsc/powershell.rb
ferventcoder: No, that's good enough. :)

lib/puppet_x/puppetlabs/powershell_manager.rb
Iristyle: The line on 138 blocks until the `@output_ready_event` is signaled from the PowerShell side after it has executed and written all it's data (the PS code is completely synchronous).

By the time we reach line 141, I'm not sure we even need to verify with `IO.select` that the pipe is readable (as the PowerShell code has completely executed by then), but I added it as an additional safety measure to prevent an attempt to read the pipe prior to the output being flushed by Ruby.

joshcooper: > line on 138 blocks until the @output_ready_event is signaled from the PowerShell side after it has executed and written all it's data

If powershell needs to write more than the named pipe buffer (I forget if it's 4 or 8K by default), then ruby and powershell will deadlock. That's because powershell will block writing to stdout waiting for ruby to drain the pipe, and ruby will wait for powershell to signal the ready event, which will never happen.

Iristyle: Ah, I understand what you're saying now.  I will verify that.

FWIW, the PowerShell code will never write even close to 4k.  Even on erroneous runs, we're looking at maybe 512 bytes.  Most runs return a < 100 character block of JSON, as the PS code is executed within a `try` / `catch`

Iristyle: I've verified that flushing up to 32k from the PowerShell side works OK.  We hit the timeout writing 64k in a single execution via:

```powershell
$bytes_in_k = 1024 * 64
[Text.Encoding]::UTF8.GetString((New-Object Byte[] ($bytes_in_k))) | Write-Output
```
I backed off the 64k with `$bytes_in_k = 1024 * 64 - 1024` (assuming that stdout had some other data written along the way), and that didn't timeout.

So it would appear the buffer here is 64k.  I can make some minor modifications to the code to prevent the race as you've suggested.


Thanks @joshcooper !

jpogran:
Passes acceptance

       Total Suite Time: 1967.78 seconds
      Average Test Time: 29.37 seconds
              Attempted: 67
                 Passed: 67
                 Failed: 0
                Errored: 0
                Skipped: 0
                Pending: 0
                  Total: 67

      - Specific Test Case Status -

ferventcoder:
@jpogran @Iristyle does this supersede #165? If so I think we should close one so we can move forward without confusion.

ferventcoder:
Or does this depend on that PR being merged first?

Iristyle:
I'm pulling in bits from this PR into #165

Iristyle:
Closed in favor of #165

PR #169, opened 2016-01-07
(maint) Update Dependencies to use / separator

ferventcoder:
In older versions of Puppet, there is a bug where a dependency is not
realized as installed when using "-" e.g. "puppetlabs-reboot" versus
"puppetlabs/reboot". Update to using "/" to guarantee compatibility
with older versions of Puppet.

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -37,15 +37,15 @@
   ],
   "dependencies": [
     {
-      "name": "puppetlabs-stdlib",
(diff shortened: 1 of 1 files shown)


Iristyle:
:+1: but do we really want a PUP ticket number as the subject line in a module commit?

ferventcoder:
I can switch it to maint.

ferventcoder:
And now it is maint.

PR #172, opened 2016-01-13
(MODULES-2709) Implement DSC timeout inside PS

Iristyle:
- Previously, run timeouts were executed on the Ruby side, which had
   the side effect of allowing PowerShell code to continue to run in
   the background, which could lead to a host of timing problems.

   For instance, a long-running DSC invocation that entered a timeout
   state on the Ruby execution side, would efectively block any
   additional PowerShell code from running until it had completed.

   Rework the timeout to take effect on the PowerShell side.  Create
   runspaces as needed (when one does not exist, or when the previous
   runspace encountered a timeout and was forcefully Disposed of).
   Otherwise, reuse runspaces for performance to behave similarly to
   calling ScriptBlock.invoke().

   This should clear up any testing issues that have cropped up as a
   result of previous runs still executing that were timeout'd.


ferventcoder:
Waiting on builds now.

PR #177, opened 2016-01-15
(MODULES-2772) Unpin master_manipulator

cowofevil:
There was a bug in master_manipulator that has been fixed which caused test
failures on PE 3.8.x. The master_manipulator gem was pinned to a known working
version which is no longer necessary.


ferventcoder:
LGTM

PR #179, opened 2016-02-03
(FM-4661) Improve DSC Build Import Process

glennsarti:
Previously users were having issues when importing custom written DSC
modules.  This was mainly due to unexpected module directory structure and
unhelpful error messages from the build process.

- Added more information when failing to find the required DSCResources
  subdirectory
- Added a warning when the module name is detected as dsc_resources.  This
  can happen if a user imports the module too high up the directory structure
- Improved the build readme with a better picture of where in the directory
  structure the import should be started


jpogran:
Image has a typo. Should be 'and' not 'an'

build/dsc/resource.rb
```diff
@@ -116,13 +116,18 @@ def has_credential?
...
         path_array = @resource_mof_path.split('/')
         revert_array = path_array.reverse
         downcased_array = revert_array.collect{|p| p.downcase}
         index = downcased_array.index('dscresources')
-        raise "module for #{self.name} not found" if index == nil
+        raise "module for #{self.name} not found (Missing DSCResources directory in path #{@resource_mof_path})" if index == nil
 
         module_name = revert_array[index + 1 ] rescue nil
+        if (module_name == "dsc_resources") 
+          warn("The name of the module in directory #{@resource_mof_path} was detected as #{module_name}.  This may indicate that the wrong directory was used to import the resource")
```
Iristyle: If we know the bad dir is `dsc_resources`, do we need to interpolate it?

README_BUILD.md
```diff
@@ -53,17 +53,33 @@ build your types.
...
 
 ### Steps to Build
 
 When importing or creating custom types, follow these steps:
 
 1. Build the module with `DSC_REF=84a467c bundle exec rake dsc:build`, where `84a467c` is the SHA of the repository https://github.com/PowerShell/DscResources that the current Puppet types are based on.
-2. Now take your own modules path and import your types: `bundle exec rake dsc:resources:import["path/to/your/types"]`. This should be the parent path that contains a folder (or folders) of DSC Resources. e.g. in the example above, the MyModule could be a folder called MyDscResources - you would run `bundle exec rake dsc:resources:import["build/vendor/custom/MyDscResources"]`.
+2. Now take your own modules path and import your types: `bundle exec rake dsc:resources:import["path/to/your/types"]`. This should be the parent path that contains a folder (or folders) of DSC Resources.
+   e.g. run `bundle exec rake dsc:resources:import["build/vendor/custom"]`.  
+```
```
Iristyle: :+1: this is helpful

Iristyle:
The image is from @ferventcoder @jpogran

README_BUILD.md
ferventcoder: :+1:

build/dsc/resource.rb
glennsarti: I wanted to keep this PR fairly simple and note create additional functionality yet i.e. docs and error message improvements.  Ideally, the import process should be doing this detection, not the build process (fail fast).  I will raise an additional ticket to review the import/build process scripts to take into account more varied directory structures e.g. Right now, it is difficult to import a single StackExchange resource

(https://github.com/PowerShellOrg/StackExchangeResources/tree/master/DSCResources)

ferventcoder:
LGTM :+1:

build/dsc/resource.rb
ferventcoder: I think @Iristyle is suggesting that because `if (module_name == 'dsc_resources')` already, do you really need to interpolate that variable in the warning because it would only be equal to `dsc_resources` currently. It's a micro optimization as a review point.

ferventcoder: I think it is fine because if the if statement is ever changed, then it works without changes. Otherwise someone needs to remember to update this as well.

glennsarti: My mistake.  @ferventcoder, your interpretation is why I used the variable name, not the explicit text.

PR #181, opened 2016-03-14
(MODULES-3132) Update MOF Parser to modify files with a BOM

glennsarti:
The MOF parser used in the DSC module does not allow BOM (Byte Order Mark) in the text file.  This
can be introduced by third parties as this module depends on the upstream Powershell Github
repository.  This commit adds two steps to the rake rasks.  The first is a new rake task called
dsc:sanitize.  This will parse all MOF files and remove BOM markings if found.  The second is an
additional step to the build process which will scan all MOF files for BOM markings and fail early
if found.

Also in this commit the example SHA in the README_BUILD.md file is incremented to the latest
Powershell GitHub SHA.  This is required as the previous SHA did not include any files with a BOM.

### DIFF
--- a/build/dsc.rake
+++ b/build/dsc.rake
@@ -24,9 +24,72 @@ namespace :dsc do
     end
 
     Rake::Task['dsc:resources:import'].invoke unless File.exists?(default_dsc_resources_path)
+    raise "The MOF files are not sanitized" unless sanitize_mof_files("#{default_dsc_resources_path}/**/*.mof", false)    
     Rake::Task['dsc:types:clean'].invoke(dsc_module_path)
     Rake::Task['dsc:types:build'].invoke(dsc_module_path)
     Rake::Task['dsc:types:document'].invoke(dsc_module_path)
+  end
+
+  def sanitize_mof_files(path_spec, remediate = false)
+    issanitized = true
+    detector = CharlockHolmes::EncodingDetector.new
+    Dir[path_spec].each do |original|
+      if !File.directory?(original)
+        content = File.read(original)
+        detection = detector.detect(content)
+        if detection[:encoding] != 'ISO-8859-1'
+          # Check if a BOM is present
+          bomtype = nil
+          bombytes = IO.binread(original,4)
+          case bombytes[0..3]
+            when "\x00\x00\xFE\xFF"
+              bomtype = bomtype || 'BOM_UTF_32_1'
+            when "\xFF\xFE\x00\x00"
+              bomtype = bomtype || 'BOM_UTF_32_2'
+            when "\xDD\x73\x66\x73"
+              bomtype = bomtype || 'BOM_UTF_EBCDIC'
+            when "\x84\x31\x95\x33"
+              bomtype = bomtype || 'BOM_GB_18030'
+          end
+          case bombytes[0..2]
+            when "\xEF\xBB\xBF"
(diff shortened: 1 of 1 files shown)


Iristyle:
Can probably squash this down @glennsarti and if you want to add a comment that says `paired-with: James Pogran <james.pogran@puppetlabs.com>` that's probably fine.

Also, minor nit that it looks like you're exceeding 72 chars in the commit msg body, so it's rendering a bit funny (on GitHub and in terminal)

glennsarti:
Commits were squashed and commit message ammended

Iristyle:
So in looking at this in more detail... wondering why you chose to go the rake route, rather than modifying the MOF parser code?

I opened up a PR showing the mods necessary to the MOF parser at https://github.com/puppetlabs/mof/pull/2.  Advantages:

* Less code changed (though admittedly it's a bit ugly)
* Parsing not performed in multiple locations
* Doesn't pull Charlock into our Rakefile (which we know won't work on Windows)
* Doesn't rewrite files (which could be considered undesirable)

In some rudimentary testing, the results appear to be the same (still could use some additional validation though).

glennsarti:
I chose the Rake route to allow a user to independently verify the the MOF files prior to import i.e. fail fast.  Also by rewriting the file, a user would be able to verify the content of the file prior to import, if they wished e.g. What if the conversion process munged the file somehow?  If it's inline to the parsing then it's very difficult to diagnose that the content has changed.

That being said I can definitely see the advantages in using your suggested method.  I'll do a quick spike based on your PR and extended it to all possible BOMs and see if I can get the best of both ideas.

Iristyle:
FWIW, I think all possible BOMs is probably overkill.  I would surprised to see anything other than UTF-16LE and UTF-8.  

Sublime can handle:

* `UTF-8 with BOM`
* `UTF-16LE with BOM`
* `UTF-16BE with BOM`

Notepad++ can handle:

* `UTF-8 with BOM`
* `UCS-2 LE with BOM`
* `UCS-2 BE with BOM`

Of course, `UCS-2` is a cousin to `UTF-16`, so I believe they are actually the same.  All of the other esoteric variations are ... well... esoteric.  I also wonder if any of their characters appear in legitimate content for the other encoding types (which could be problematic for us)

Iristyle:
Also - I think your rake idea is useful, but maybe as a separate task outside of the normal workflow.  The MOF parser does currently tell you where it died (I have an incomplete branch that perhaps I should revive that will actually tell you all bad files rather than first bad file).  So I wouldn't be opposed to enabling the ability to file mangle up-front to rake tasks for the debugging reasons you describe.

However, I think there are a few things that make that less desireable in this PR:

* Encoding handling in multiple spots (I think that either rake OR lexer / parser would be best)
* If encoding handling is moved from the MOF parser, to external tasks, it is in turn less resilient as it now relies on the presence of the rake tasks to fix files for it (usually handling of encodings would be performed by a lexer / parser)
* I'll reiterate the Charlock Holmes / Windows compatibility point as well (would love to remove it from the process completely)

I think in an ideal world, the MOF parser library handles everything encoding related... and if we want to enable this rake task based workflow, there are would have some public helpers that the rake tasks could call.  Of course, that's a bit out of the scope of the current efforts here ;0

Iristyle:
After pairing with @glennsarti, closing in favor of https://github.com/puppetlabs/mof/pull/2

PR #182, opened 2016-03-14
(MODULES-3119) Remove unncessary files from module build

glennsarti:
This commit adds extra directories to the PMT Ignore file as, build, docs and tests are
not required at runtime for DSC module.  Also added *.local to ignore local development
files.  While the addition of these files will not cause any errors, they are removed
to keep the module installation as lean as possible.

### DIFF
--- a/.pmtignore
+++ b/.pmtignore
@@ -2,5 +2,14 @@ import/
 /spec/fixtures/
 .tmp
 *.lock
+*.local
 .rbenv-gemsets
 .ruby-version
+build/
+docs/
+tests/
+
+# Normally the spec directory is included in module distribution, however due to large number of 
+# files generated, they are ignored when packaging the module.  The spec tests are not required
+# for the module to operate
+spec/


ferventcoder:
Looks legit to me.

.pmtignore
```diff
@@ -2,5 +2,9 @@ import/
 /spec/fixtures/
 .tmp
 *.lock
+*.local
 .rbenv-gemsets
 .ruby-version
+build/
+docs/
+tests/
```
ferventcoder: While we are at it, we could probably add the specs directory.

ferventcoder:
Wow, 778KB versus the 8-ish MB of the last release.

.pmtignore
glennsarti: spec added with reasoning as to why as it's not normally excluded

ferventcoder:
I'm a :+1: on this to merge.

Iristyle:
* `.tar.gz` from `1751K` down to `646K`
* embedded `.tar` from `14,290K` down to `6,200K`
* `2602 files` / `460 folders` down to `805 files` / `344 folders`

Extracted and compared differences using BeyondCompare ... nothing extra was cut out.  :+1:

PR #183, opened 2016-03-16
(MAINT) Update Gemfile and spec test value generation

glennsarti:
This commit adds a SHA reference to the MOF Gem that is retrieved from
git.  The MOF repoisitory has no release tags or branches so a commit
SHA is used.

This commit also adds spec test values for the real32 and real64
datatypes.  These datatypes are floating point numbers represented as
strings.  These were missing and caused spec unexpected test failures.


Gemfile
```diff
@@ -54,7 +54,7 @@ end
 
 group :build do
   gem 'cim',                :require => false
-  gem 'mof',                :require => false, :git => 'https://github.com/puppetlabs/mof.git'
+  gem 'mof',                :require => false, :git => 'https://github.com/puppetlabs/mof.git', :branch => '2f42f1c'
```
Iristyle: I always thought that bundler used `:ref` for this... are `:ref` and `:branch` interchangeable?

Should probably double check Travis / AppVeyor to ensure we pulled the right version.  Should also probably wait until we merge h the new changes to MOF repo in before we merge this.

ferventcoder:
The sync.yaml also needs hit, although this is likely just temporary. I'm :+1:

PR #184, opened 2016-03-17
(MODULES-3179) Add DSC Reference to the module when importing types

glennsarti:
This commit addes the DSC_REF environment variable to the types.md
file to help signify which git reference was used during the build
process.  This does not affect the operation of the module or the
DSC resources, but it's used for documentation purposes only.

example types.md;
```
...
...
##### xWordPress

Puppet Type | DSC Resource | Github Repo
----------- | ----------------- | -----
dsc_xwordpresssite | [xWordPressSite](https://github.com/puppetlabs/puppetlabs-dsc/tree/master/lib/puppet_x/dsc_resources/xWordPress/DscResources/MSFT_xWordPressSite) | [repo](https://github.com/PowerShell/xWordPress)


Git reference for DSC: 84a467c
```

### DIFF
--- a/build/dsc/manager.rb
+++ b/build/dsc/manager.rb
@@ -12,6 +12,7 @@ module Dsc
 
       @import_folder            = "#{@module_path}/import"
       @dsc_modules_folder       = "#{@import_folder}/dsc_resources"
+      @dsc_git_reference        = "#{@import_folder}/dsc_resources/dsc_reference.txt"
       @dmtf_mof_folder          = "#{@module_path}/build/vendor/dmtf_mof"
 
       @type_template_file       = "#{@dsc_lib_path}/templates/dsc_type.rb.erb"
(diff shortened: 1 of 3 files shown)


glennsarti:
I looked at dropping a text file in the build process and other options, but modifying the types.md seemed the easiest to do and so

build/dsc/manager.rb
```diff
@@ -188,6 +188,8 @@ def document_types(markdown_file_path, dsc_types)
           end
           file.write("\n")
         end
+        
+        file.write("\nGit reference of DSC used to create types: #{ENV['DSC_REF']}")
```
Iristyle: FYI - `DSC_REF` is not something that must be specified to perform a `dsc:build` ... so you might want to actually just do a `git rev-parse --short HEAD` against the import directory instead.

glennsarti: The import directory is purged of git history as part of https://github.com/puppetlabs/puppetlabs-dsc/blob/master/build/dsc.rake#L70 .  I can modify it to only add that footnote if `ENV['DSC_REF']` is defined.

Iristyle: If we're going to use an env var to store that global state, then I think we could set it after the repo is cloned if it hasn't already been set?  (I know this won't work always based on cached files, but should work some of the time).

glennsarti:
@Iristyle Latest commit detects the git reference from the repository and then stores the ref as a text file in the import directory.  During the `dsc:types:document` task, this reference is appended to the types.md files.  Added two helper methods to the manager to support reading and writing the reference.

build/dsc/manager.rb
```diff
@@ -210,9 +211,24 @@ def document_types(markdown_file_path, dsc_types)
           end
           file.write("\n")
         end
+        
+        gitref = get_powershell_dsc_git_reference        
+        file.write("\nThe vendored Powershell DSC Resource can be rebuilt using `DSC_REF=#{gitref}`") if gitref
```
jpogran: I would prefer to see this at beginning of type file. if we put it at the end, then each time we get a new DSC Resource we move this line.

glennsarti:
Closing as no longer required

PR #187, opened 2016-03-17
(MODULES-3133) Re-vendor DSC resources

Iristyle:


build/dsc/manager.rb
```diff
@@ -214,6 +214,41 @@ def clean_dmtf_mofs
       clean_folder([@dmtf_mof_folder])
     end
 
+    def format_embeddedtype_all_values(property)
```
Iristyle: This isn't quite right as it fails when `manager.rb` has the extraneous types omitted.

Iristyle:
@glennsarti pending another look by you, this should be good..

build/dsc/templates/dsc_type.rb.erb
```diff
@@ -108,7 +108,7 @@ Puppet::Type.newtype(:dsc_<%= resource.friendlyname.downcase %>) do
           .select { |c| c[:klass].name == property.embeddedinstance_class_name }
           .first[:klass]
           .features
-          .map { |p| Dsc::Property.new(p) }
+          .map { |p| Dsc::Property.new(p, resource.relative_mof_path) }
```
ferventcoder: Interesting.

build/dsc/resource.rb
```diff
@@ -16,6 +16,7 @@ def initialize(mof_class, mof_path)
       @embedded_properties = nil
       @dsc_module          = nil
       @ps_module           = nil
+      @absentable          = nil
```
ferventcoder: Did I miss something? I thought we already had this...?

build/dsc/manager.rb
```diff
@@ -32,9 +32,6 @@ def initialize
         'MSFT_Credential' => { 'user' => 'user', 'password' => 'password' },
         'MSFT_KeyValuePair' => { 'somekey' => 'somevalue' },
         'MSFT_KeyValuePair[]' => { 'somekey' => 'somevalue', 'somekey2' => 'somevalue2' },
-        # these hashes will be automatically wrapped in tests
-        'MSFT_xFileDirectory[]' => { 'DestinationPath' => 'c:/foo/bar', 'Recurse' => true },
-        'MSFT_xWebBindingInformation[]' => { 'Port' => 8080, 'Protocol' => 'https' },
```
ferventcoder: We were able to remove these?

ferventcoder:
On the whole I'm +1 with what I can see. I imagine that the vendored files have changed as well.

build/dsc/resource.rb
Iristyle: That's about setting a default so we can do a simple comparison to see if it's been calculated already (as we query `absentable` quite a bit when building templates)

build/dsc/manager.rb
Iristyle: Last commits generates EmbeddedInstance spec values on the fly... no more need to statically define these.

glennsarti:
@Iristyle Did you want to update the README_BUILD.md file with which DSC SHA reference you used? (0a9d28f) or leave that for MODULES-3179 to take care of.

build/dsc/templates/dsc_type.rb.erb
glennsarti: Without this being passed through, the Property can only emit error information with only the property name e.g. 'Property 'Ensure' is not valid'.  This is not useful during debugging as it does not tell you which of the 200+ MOF files have the property, particularly in our case which was debugging the Ensure property.  By passing the full path to the schema MOF file, useful error messages can be generated.

Iristyle:
Thanks for catching that @glennsarti - updated `README` and repushed.

build/dsc/templates/dsc_type.rb.erb
ferventcoder: :+1:

build/dsc/manager.rb
ferventcoder: :+1:

PR #188, opened 2016-03-17
(maint) Simple batch file to generate tarballs with good long path names

ferventcoder:


ferventcoder:
closed to retarget to a different branch.

PR #190, opened 2016-03-25
(FM-5017) Update Puppet Agent for Acceptance Tests

cowofevil:
The acceptance tests were hardcoded to run against 1.3.2 to work-around a
Beaker limitation. The newer versions of Beaker now support installing the
latest Puppet Agent without specifying a version number. Updated the pre-suite
for acceptance tests to now automatically grab the latest version.

### DIFF
--- a/tests/acceptance/pre-suite/01_puppet_agent_install.rb
+++ b/tests/acceptance/pre-suite/01_puppet_agent_install.rb
@@ -2,11 +2,12 @@ test_name 'Install Puppet Agent'
 
 confine(:to, :platform => 'windows')
 
-#Init
-puppet_agent_version = ENV['BEAKER_PUPPET_AGENT_VERSION'] || '1.3.2'
-
 step 'Install Puppet Agent'
-install_puppet_agent_on(agents, :version => puppet_agent_version)
+if ENV['BEAKER_PUPPET_AGENT_VERSION']
+  install_puppet_agent_on(agents, :version => ENV['BEAKER_PUPPET_AGENT_VERSION'])
+else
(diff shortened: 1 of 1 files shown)


tests/acceptance/pre-suite/01_puppet_agent_install.rb
```diff
@@ -2,11 +2,12 @@
...
-#Init
-puppet_agent_version = ENV['BEAKER_PUPPET_AGENT_VERSION'] || '1.3.2'
-
 step 'Install Puppet Agent'
-install_puppet_agent_on(agents, :version => puppet_agent_version)
+if ENV['BEAKER_PUPPET_AGENT_VERSION']
+  install_puppet_agent_on(agents, :version => ENV['BEAKER_PUPPET_AGENT_VERSION'])
+else
+  install_puppet_agent_on(agents)
+end
```
glennsarti: Would it be better to use the ternary if statement?
```
ENV['BEAKER_PUPPET_AGENT_VERSION'] ? install_puppet_agent_on(agents, :version => ENV['BEAKER_PUPPET_AGENT_VERSION']) : install_puppet_agent_on(agents)
```

cowofevil: I would argue that the one-liner is less readable. (From a non-Ruby person viewpoint) The if block is verbose. Also the one-liner is 151 characters long meaning that for some people that scrolls off to the right of the screen which decreases readability.

I'm not passionate about it and could easily be swayed to use the one-liner. However, I don't see what value it adds other than it uses less lines of code.

PR #194, opened 2016-04-22
(MODULES-3288) Multi-line values do not work in DSCResources

jakeballard:
Enable multi-line DSC values. 

It uses Powershell here string formatting when the value given has multiple lines. Single line strings are still formatted inside single quotes. This required changing the Powershell manager to use a code block instead of a here string for storing the script to run, because you cannot nest a single quoted here string inside of another.

### DIFF
--- a/lib/puppet/provider/base_dsc/powershell.rb
+++ b/lib/puppet/provider/base_dsc/powershell.rb
@@ -143,6 +143,8 @@ Puppet (including 3.x), or to a Puppet version newer than 3.x.
 
   def self.format_dsc_value(dsc_value)
     case
+    when dsc_value.class.name == 'String' && dsc_value.lines.count > 1
+      "@'\n#{dsc_value}\n'@"
     when dsc_value.class.name == 'String'
       "'#{escape_quotes(dsc_value)}'"
     when dsc_value.class.ancestors.include?(Numeric)
--- a/lib/puppet_x/puppetlabs/powershell_manager.rb
+++ b/lib/puppet_x/puppetlabs/powershell_manager.rb
@@ -35,9 +35,9 @@ module PuppetX
           $runspace.Open()
         }
 
-        $powershell_code = @'
+        $powershell_code = {
 #{powershell_code}
-'@
+}
         $ps = $null
 
         try
--- a/lib/puppet_x/puppetlabs/powershell_manager.rb
+++ b/lib/puppet_x/puppetlabs/powershell_manager.rb
@@ -1,6 +1,7 @@
 require 'securerandom'
 require 'open3'
 require 'ffi'
+require 'base64'
 
 module PuppetX
   module Dsc
@@ -25,6 +26,7 @@ module PuppetX
       def execute(powershell_code, timeout_ms = 300 * 1000)
         output_ready_event_name =  "Global\\#{SecureRandom.uuid}"
         output_ready_event = self.class.create_event(output_ready_event_name)
+        powershell_code_base64 = Base64.encode64(powershell_code)
 
         # always need a trailing newline to ensure PowerShell parses code
         code = <<-CODE
@@ -35,9 +37,10 @@ module PuppetX
           $runspace.Open()
         }
 
-        $powershell_code = {
-#{powershell_code}
-}
+        $powershell_code_base64 = '#{powershell_code_base64}'
+        $powershell_code_bytes  = [System.Convert]::FromBase64String($powershell_code_base64)
+        $powershell_code = [System.Text.Encoding]::UTF8.GetString($powershell_code_bytes)
+
         $ps = $null
 
         try
--- a/spec/integration/puppet_x/puppetlabs/powershell_manager_spec.rb
+++ b/spec/integration/puppet_x/puppetlabs/powershell_manager_spec.rb
@@ -43,6 +43,18 @@ $count
       expect(result).not_to eq(nil)
     end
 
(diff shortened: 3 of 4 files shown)


kenazk:
Thanks for contributing! Please see the guide here for contributions, including detailed commit messages, and running the tests. 

https://docs.puppet.com/forge/contributing.html

For example, the Powershell manager seems to be complaining about changing the fundamental type from string to code block.

jakeballard:
Thanks for the feedback. I'll fix the commit message, I was expecting the pull request to make a new commit the comment. 

I'll need some help on the tests. When I ran `rake spec` on my system it returned without failures. Is there something else I need to run? 

```
Finished in 1 minute 17.99 seconds (files took 22.91 seconds to load)
18384 examples, 0 failures
```

jakeballard:
I was able to figure out the tests, and saw the same failure you mentioned. I've updated it to pass the code as a base64 encoded string instead, and the tests were then able to pass. I also added tests for the changes.

I hope I got the new commit messages right. However I'm not sure how to modify the original ones, apart from recreating it on a new branch.

jpogran:
@jakeballard Thanks for your work. We are working towards an imminent release and are evaluating this PR for that release. Thanks for your patience while we work through this.

With regards to the commits, what we are looking for is the latest 2 commits be squashed into the respective first commits during a rebase. You should end up with only the last two commits and their current commit messages when you are done. If you are uncomfortable with doing this we can either walk you through it or handle it ourselves if we merge this PR.

jakeballard:
Rebasing the commits shouldn't be a problem, I just didn't realize that was an option, all the docs said not to change history on things already pushed to a public repo. 

The DSC resource I found that needed multi-line values was dsc_xjeatoolkit, one of the values wants a CSV file as a string. Like this:

```
dsc_xjeatoolkit {'Service':
    dsc_name         => 'Service',
    dsc_commandspecs => @("SPEC"/n$)
      Name,Parameter,ValidateSet,ValidatePattern
      Get-Service
      |-SPEC
  }
```

However the resource itself didn't seem to need any code changes, it was happy being given a here string. The problem is the formatter still treats it as a normal string and sends it to Powershell in single quotes. Like this:

```
commandspec = 'Name,Parameter,ValidateSet,ValidatePattern
Get-Service'
```

but it needs to be a here string like this:

```
commandspec = @'
Name,Parameter,ValidateSet,ValidatePattern
Get-Service
'@
```
Then the Powershell manager was already using a single quoted here string to hold the DSC code, and you can't nest a single quoted here string inside of another. So I changed that to instead pass it as a base64 encoded string, which Powershell decodes.

jpogran:
Thanks for putting the example back. I tested this out in puppet 4 using a heredoc like so:

```
dsc_xjeatoolkit{'process':
    dsc_name         => 'Process',
    dsc_commandspecs => @(END)
Name,Parameter,ValidateSet,ValidatePattern
Get-Process
Get-Service
Stop-Process,Name,calc;notepad
Restart-Service,Name,,^A
END
}
```

This produces the following DSC code:

```
$invokeParams = @{
  Name          = 'xJeaToolKit'
  Method        = 'set'
  Property      = @{
    name = 'Process'
    commandspecs = 'Name,Parameter,ValidateSet,ValidatePattern
Get-Process
Get-Service
Stop-Process,Name,calc;notepad
Restart-Service,Name,,^A
'
  }
  ModuleName = @{
    ModuleName      = "C:/foo/modules/dsc/lib/puppet_x/dsc_resources/xJea/xJea.psd1"
    RequiredVersion = "0.3.0.0"
  }
}
```

Which runs correctly and generates the following files:

```
PS C:\foo\modules\dsc> ls $env:ProgramFiles\Jea\Toolkit

    Directory: C:\Program Files\Jea\Toolkit

Mode                LastWriteTime         Length Name
----                -------------         ------ ----
-a----        5/13/2016   8:44 AM            252 Process-CommandSpecs.csv
-a----        5/13/2016   8:44 AM          16802 Process-ToolKit.psm1
```

I copied the DSC code and ran it interactively as well, and it produces the correct data:

```
PS C:\foo\modules\dsc> $invokeParams

Name                           Value
----                           -----
Method                         set
Property                       {name, commandspecs}
Name                           xJeaToolKit
ModuleName                     {RequiredVersion, ModuleName}

PS C:\foo\modules\dsc> $invokeParams.Property

Name                           Value
----                           -----
name                           Process
commandspecs                   Name,Parameter,ValidateSet,ValidatePattern...

PS C:\foo\modules\dsc> $invokeParams.Property.commandspecs
Name,Parameter,ValidateSet,ValidatePattern
Get-Process
Get-Service
Stop-Process,Name,calc;notepad
Restart-Service,Name,,^A
```

If you look at the MOF schema file https://github.com/PowerShell/xJea/blob/dev/DSCResources/MSFT_xJeaToolkit/MSFT_xJeaToolkit.schema.mof, you can see the CommandSpec property is a string. There isn't a requirement to make it a PowerShell here string, thats just a PowerShell language construct. The end result that is necessary is a string with newlines in it.

Please let me know if this explanation helps. If it doesn't please open a MODULES ticket and we'll get it prioritized.

jakeballard:
You are right, I just retried my Puppet code without these changes and it's working. Not sure now what I did wrong originally, dsc_xjeaendpoint wouldn't apply the toolkit files it was outputting, but I must have fixed something else in my Puppet code while troubleshooting this and misattributed.

Sorry about that. I'll close this.

PR #199, opened 2016-05-10
(MODULES-3343) Nano server compatibility

Iristyle:
- Use the CoreCLR compatible `WaitHandle.WaitOne(timeout)`
- Don't invoke `powershell.exe` with `-Sta`


jpogran:
👍

PR #201, opened 2016-05-13
(MODULES-3342) Increase execution timeout

jpogran:
Increase the timeout to 20 minutes for all DSC executions


ferventcoder:
Since this changed in 3 places, I would see you create one constant and change these values to that constant. Then put a comment above it stating that the value represents 20 minutes.

PR #202, opened 2016-05-13
(MODULES-3363) Support new values for Ensure

jpogran:
Some DSC Resources use arbitrary values for Ensure or Absent. The
xRunbookDirectory DSC Resource in the xSCSMA module uses Published
and Draft as valid values for Ensure. This fix adds this as valid
values


build/dsc/resource.rb
```diff
@@ -110,7 +110,12 @@ def ensure_value
           'present'
         elsif ensure_property.values.any? { |v| v.casecmp('enable') == 0 }
           'enable'
+        elsif ensure_property.values.any? { |v| v.casecmp('published') == 0 }
+          'published'
+        elsif ensure_property.values.any? { |v| v.casecmp('draft') == 0 }
+          'draft'
         else
+          require "pry";binding.pry
```
Iristyle: @jpogran looks like we have a straggler ;0

build/dsc/resource.rb
```diff
@@ -110,7 +110,12 @@ def ensure_value
           'present'
         elsif ensure_property.values.any? { |v| v.casecmp('enable') == 0 }
           'enable'
+        elsif ensure_property.values.any? { |v| v.casecmp('published') == 0 }
```
Iristyle: Ok, so looking at https://github.com/PowerShell/xSCSMA/blob/dev/DSCResources/MSFT_xRunbookDirectory/MSFT_xRunbookDirectory.schema.mof#L4 I see the basis for this PR.

However, I think what you want to do here is treat `published` OR `draft` as a valid ensure value (at runtime).  However, the change here applies to the type generation (not looking at the resource at runtime) - so this only takes effect for the MOF schema parsing / type generation.

I think what you'll find, when you generate the types, is that you will only see `published` show up as an allowed `ensure` value and if you tried to use `draft`, it wouldn't do anyting.  You need to inform the Puppet side that either of these values are valid values for `ensure`, and that they should trigger enforcement behavior in Puppet.

I'm not sure how well that is going to work, but try it and find out once you're able to update this PR and include the generated types here.  I think it might be a little wonky since we've built our mapping layer around just the concepts of `ensure` / `absent`, but not any third state.

jpogran: Does any of the above change if `Published` and `Draft` are treated as `Present` inside the DSC Resource code? Looking through the [code](https://github.com/PowerShell/xSCSMA/blob/dev/DSCResources/MSFT_xRunbookDirectory/MSFT_xRunbookDirectory.psm1#L155-L200), it's either if `Absent` remove **OR** ensure it's `Present` and set the state of the Runbook as `Published` or `Draft`. In essence they both mean its `Present`.

My instinct is that this is abuse of the Ensure parameter by the DSC Resource and that this should be a Ensure and a State parameter, but they moved from that implementation to this so it's unlikely they would change back.

build/dsc/resource.rb
```diff
@@ -93,26 +93,29 @@ def absentable?
...
-        if ensure_property.values.any? { |v| v.casecmp('present') == 0 }
-          'present'
-        elsif ensure_property.values.any? { |v| v.casecmp('enable') == 0 }
-          'enable'
-        else
-          throw 'Error processing MOF schema - could not determine equivalent \'present\' value for ensure'
-        end
+    def valid_ensure_present_values
+      foo = ensure_hash.detect{|v| v[:name] == name }
+      foo != nil ? foo[:present] : ['present']
```
jpogran: This doesn't throw anymore. Think the whole approach would need to change in order to do so

build/dsc/resource.rb
```diff
@@ -93,26 +93,29 @@ def absentable?
...
-    def absent_value
-      @absent_value ||=
-        if ensure_property.values.any? { |v| v.casecmp('absent') == 0 }
-          'absent'
-        elsif ensure_property.values.any? { |v| v.casecmp('disable') == 0 }
-          'disable'
-        else
-          throw 'Error processing MOF schema - could not determine equivalent \'absent\' value for ensure'
-        end
+    def ensure_hash
```
Iristyle: nit - this returns an array, but states that it's a hash.  I think it might be better to index by name like:

```ruby
{
  'MSFT_xRunbookDirectory' => { :present => ['published', 'draft'], :absent => [ 'absent'] },
  'MSFT_WindowsOptionalFeature' => { :present => ['enable'], :absent => [ 'disable'] }
}
```

build/dsc/resource.rb
```diff
@@ -93,26 +93,29 @@ def absentable?
...
-      @ensure_value ||=
-        if ensure_property.values.any? { |v| v.casecmp('present') == 0 }
-          'present'
-        elsif ensure_property.values.any? { |v| v.casecmp('enable') == 0 }
-          'enable'
-        else
-          throw 'Error processing MOF schema - could not determine equivalent \'present\' value for ensure'
-        end
+    def valid_ensure_present_values
+      foo = ensure_hash.detect{|v| v[:name] == name }
```
Iristyle: Then this can just be `ensure_hash[name]` (unless we want to do case insensitive, in which case I suggest we store lower case versions in the hash and call `name.downcase` here for a key)

build/dsc/resource.rb
Iristyle: You can get to where you need to go with a few tweaks here

build/dsc/templates/dsc_type.rb.erb
```diff
@@ -86,11 +86,15 @@ Puppet::Type.newtype(:dsc_<%= resource.friendlyname.downcase %>) do
...
   ensurable do
     newvalue(:exists?) { provider.exists? }
-    newvalue(:<%= begin resource.ensure_value rescue 'present' end %>) { provider.create }
+<%  resource.valid_ensure_present_values.each do |property| -%>
+    newvalue(:<%= property.downcase %>) { provider.create }
+<%  end -%>
 <%  if resource.absentable? -%>
-    newvalue(:<%= resource.absent_value %>)  { provider.destroy }
+<%  resource.valid_ensure_absent_values.each do |property| -%>
+    newvalue(:<%= property.downcase %>)  { provider.destroy }
```
Iristyle: Hopefully we never see multiple `absent` states... as that doesn't really make any sense.

Different creation states can make sense sometimes, but I'm not so sure the `absent` state parallel holds true.  

Thoughts on whether we should even attempt / address this?

jpogran: Implementing it kept a certain symmetry and was easy to do, so I followed the path. If there is variance with present, it stands to reason there could variance with absent. Argument goes both, ways: future pain or present effort. Don't have a preference.

Iristyle: Yup, understand the symmetry argument - it's just that I'm not sure it's something we need to support, because I don't think there would ever be multiple values that *should* trigger `provider.destroy`.  Even with AWS EC2 instances, that have the most `ensure` values that I've seen (`present`, `absent`, `running`, `stopped`), only `absent` calls `destroy` because `stopped` has a different semantic meaning - see code at https://github.com/puppetlabs/puppetlabs-aws/blob/master/lib/puppet/type/ec2_instance.rb#L6-L31

Based on how we've implemented the bridge here, I think we can only really call `provider.create` for anything other than `absent`

Iristyle:
Don't forget that before merge we need to regenerate the types

glennsarti:
@jpogran What DSC REF did you use during the regeneration?  Perhaps that should be added to the commit message?

Iristyle:
To @glennsarti point, it's usually best to generate against the exact SHA that we've been using in the past when making changes to the codegen code.  That way we can easily see what changed in our existing types based on the modifications.

I'll fix this up when I merge / test it.

PR #204, opened 2016-05-26
(MODULES-3410) DSC fixtures uses incorrect PowerShell version

DLuCJ:
Previously, the DSC fixtures would use the latest (2.0.1) version
of Powershell instead of 1.0.6, as specified in the metadata.json.

This commit pins the powershell module to version 1.0.6 in the
fixtures.yml file as puppet_spec_helper does not correctly handle
version range strings.


Iristyle:
👍

PR #206, opened 2016-06-01
(MODULES-3363) Support arbitrary Ensure values

Iristyle:
Some DSC Resources use arbitrary values for `Present` or `Absent`. The
xRunbookDirectory DSC Resource in the xSCSMA module uses `Published`
and `Draft` as valid `Present` values for `Ensure`.

Modify the type building process to explicitly treat some resources as
exceptional cases by defining a custom hash for each MOF schema type.
This provides a way to identify non-standard resources easily, and a
process which will fail hard when new resources are introduced that
don't follow the model / aren't explicitly handled by the hash. It is
allowed to omit the :absent member of the hash for resource types
that are not absentable.

Move any inline special handling for enable / disable of the built-in
MOF type MSFT_WindowsOptionalFeature to this new hash, in addition to
MSFT_xRunbookDirectory.  This normalizes all exclusions under one
process.

Further remove any exception handling inside type / spec generation
templates, and move it out to the new helpers responsible for finding
and detecting the `Ensure` values -- raise exceptions in specific
cases that likely indicate a poorly defined / implemented MOF that
should be explicitly defined in the exclusion hash.

Types have been regenerated against the first commit after 90eba64,
which happens to be the first that includes the updated xSCSMA, which
will fail without these new type building changes:

https://github.com/PowerShell/DscResources/commit/e6974e102b8af1c4283d1d48fdce1395b949a0bd

As a result, this includes 2 new types:

dsc_runbookdirectory
dsc_smavariable

Supercedes #202

### DIFF
--- a/build/dsc/resource.rb
+++ b/build/dsc/resource.rb
@@ -87,32 +87,77 @@ module Dsc
 
     def absentable?
       if @absentable.nil?
-        @absentable = ensurable? &&
-          ensure_property.values.any? { |v| v.casecmp('absent') == 0 || v.casecmp('disable') == 0 }
+        if !ensurable?
+          @absentable = false
+        else
+          override = ensure_type_overrides[name.downcase]
+          @absentable = override ?
+            # if an override exists, the absence of the :absent hash key (or nil value at :absent) means not absentable
+            !override[:absent].nil? :
+            # otherwise use typical 'absent' heuristic
+            ensure_property.values.any? { |v| v.casecmp('absent') == 0 }
+        end
       end
       @absentable
     end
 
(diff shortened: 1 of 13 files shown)


Iristyle:
@puppetlabs/windows please have a look... there were some existing features that didn't get carried over in the work on #202 - so I reworked some of the type generation code accordingly.

Iristyle:
Do not merge until @glennsarti has completed the PR to update MODULES-3431 tests as a follow-on to #207

jpogran:
Test building with specified ref as well as from latest, both work. 👍

PR #207, opened 2016-06-01
(MODULES-3431) Fix revendor integration failure

Iristyle:
- In commit 7fb9f22e1137ea92270150b0ef1ccef73307e83e resources were
   regenerated and revendored.

   Unfortunately the dsc_xmysqlserver type changed as a result, removing
   the dsc_servicename parameter.  This invalidates the test in this
   commit.


Iristyle:
There were additional changes to the `mysql` resource that are causing CI failures - @glennsarti to prepare a new PR that temporarily skips the test as a follow up to this one.

PR #208, opened 2016-06-03
(MODULES-3431) Fix acceptance tests for MySQL DSC resource

glennsarti:
The MySQL DSC resource has recently been updated to v2.x with breaking
changes.  This has caused the DSC CI pipeline to fail.  This commit
changes the acceptance test to expect to fail.  Once the MySQL DSC
Resource authors fix their module, these tests will then fail the failure
and we can reinstate them.

This commit also modifies the presuite LCM configuration to only occur
on Windows agents as it was possible to cause test failures if run on
a master agent.


glennsarti:
Tested locally with integration tests

PR #211, opened 2016-06-29
(MODULES-3536) modsync update

glennsarti:
Module updated with lastest modsync settings


ferventcoder:
If AV is good, then 👍  for merge.

PR #218, opened 2016-07-14
(MODULES-3228) Track DSC Resources Based On Release Versions

ferventcoder:
### Use PSGallery Release Tags
Each submodule tracks on a commit SHA1, but that SHA1 doesn't
necessarily correspond or represent an actual release of that
submodule (aka DSC Resource) to the PowerShell Gallery. However,
those resources are tagged when they are released to the gallery
with *-PSGallery, where * represents the version going up to the
gallery. This allows someone to find the currently released
version of a module/DSC Resource.

Instead of trusting the tags that are available for a resource in
the list of submodules, which tends to be completely inaccurate if
you are looking for the latest release, gather all PSGallery tags
and find the latest version. Then check that version out for
building the resource.

This code is similar to what James proposed in de5a10eac0.

Paired with James Pogran <james.pogran at puppet.com>

### Track release versions
Generate and track versions being used in a file. When running, if not
updating, use the versions being tracked to ensure compatibility.

If a tracked version is not found when building, use/record the latest
version of that released resource. This allows the ability to add new
dsc resources automatically.

To update to the latest version of each resource, pass `true` as an
argument for `update_versions` to the import rake task or delete the
yaml file tracking the release versions.

### Outstanding items:

 * [x] Add a yaml to track versions
 * [x] Use the yaml by default
 * [x] `--latest` to attempt to update the versions

#### Open questions 

* [x] ~~For new items that are added or dropped, how should that be handled?~~ If the item is not already in the YAML, it will be added with the latest tag and can be adjusted manually. For items that are removed, well they will just need to be removed from the yaml eventually. That or one of the ways to update to the latest versions is to delete the yaml and let it rebuild. This will also have the effect of removing items that are no longer available. Then source control diffs will show you what has changed.


build/dsc.rake
```diff
@@ -67,16 +69,32 @@ eod
         FileUtils.cp_r "#{dsc_resources_path}/.", "#{dsc_resources_path_tmp}/"
       end
 
-      FileUtils.rm_rf(Dir["#{dsc_resources_path_tmp}/**/.git"])
```
ferventcoder: Don't delete the .git just yet.

build/dsc.rake
```diff
@@ -67,16 +69,32 @@ eod
...
 
+      puts "Getting latest release tags for DSC resources..."
+      Dir["#{dsc_resources_path_tmp}/xDSCResources/*"].each do |dsc_resource_path|
+        dsc_resource_name = Pathname.new(dsc_resource_path).basename
+        FileUtils.cd(dsc_resource_path) do
+          # --date-order probably doesn't matter
+          tags_raw = %x{ git log --tags --pretty=format:'%D' --simplify-by-decoration --date-order }
+          # if this starts to result in errors, we should explore pushing this out to a file
+          # would prefer another method over scan
+          versions = tags_raw.scan(/(\S+)\-PSGallery/).map { | ver | Gem::Version.new(ver[0]) }
```
ferventcoder: Scan is not my favorite item, so suggestions would be lovely.

build/dsc.rake
```diff
@@ -97,8 +118,8 @@ eod
         FileUtils.cp_r "#{dsc_resources_path_tmp}/.", "#{default_dsc_resources_path}/"
       end
 
-      puts "Removing extra dir #{dsc_resources_path_tmp}"
-      FileUtils.rm_rf "#{dsc_resources_path_tmp}"
+      #puts "Removing extra dir #{dsc_resources_path_tmp}"
+      #FileUtils.rm_rf "#{dsc_resources_path_tmp}"
```
ferventcoder: This is temporary

build/dsc.rake
```diff
@@ -67,16 +69,34 @@ eod
...
+          sh "git checkout #{latest_version}"
+        end
+      end
+
+      FileUtils.rm_rf(Dir["#{dsc_resources_path_tmp}/**/.git"])
+
       puts "Cleaning out test and example files for #{item_name}"
-      FileUtils.rm_rf(Dir["#{dsc_resources_path_tmp}/**/.git",
-                          "#{dsc_resources_path_tmp}/**/*[Ss]ample*",
+      FileUtils.rm_rf(Dir["#{dsc_resources_path_tmp}/**/*[Ss]ample*",
```
ferventcoder: Separated to make the git removal specific and explicit. Could put it back and change the message.

ferventcoder:
@puppetlabs/windows thoughts so far?

build/dsc.rake
```diff
@@ -55,10 +55,12 @@ eod
 
       if !is_custom_resource
         puts "Downloading and Importing #{item_name}"
-        cmd = "git clone #{dsc_repo} #{dsc_resources_path_tmp} && " +
-          "cd #{dsc_resources_path_tmp} && "
-        cmd += "git checkout #{ENV['DSC_REF']} && " if ENV['DSC_REF']
-        cmd += "git submodule update --init"
+        cmd = ''
+        # ensure we can always update what we have without requiring deletion of the folder
+        cmd = "git clone #{dsc_repo} #{dsc_resources_path_tmp} && " unless Dir.exist? dsc_resources_path_tmp
```
ferventcoder: This should probably go away since we delete this folder every time.

ferventcoder: Artifact of rapid testing, we'll say.

build/dsc.rake
```diff
@@ -55,10 +58,11 @@ eod
 
       if !is_custom_resource
         puts "Downloading and Importing #{item_name}"
-        cmd = "git clone #{dsc_repo} #{dsc_resources_path_tmp} && " +
-          "cd #{dsc_resources_path_tmp} && "
-        cmd += "git checkout #{ENV['DSC_REF']} && " if ENV['DSC_REF']
-        cmd += "git submodule update --init"
+        cmd = ''
+        cmd = "git clone #{dsc_repo} #{dsc_resources_path_tmp} && " unless Dir.exist? dsc_resources_path_tmp
```
ferventcoder: Had an earlier note that this can go away.

build/dsc.rake
```diff
@@ -67,16 +71,54 @@ eod
...
+          versions = tags_raw.scan(/(\S+)\-PSGallery/).map { | ver | Gem::Version.new(ver[0]) }
+          if versions.empty?
+            raise "#{dsc_resource_name} does not have any '*-PSGallery' tags. Appears it has not been released yet. Tags found #{tags_raw.to_s}"
+          end
+
+          latest_version = versions.max.to_s + "-PSGallery"
+          tracked_version = resource_tags["#{dsc_resource_name}"]
+
+          # placeholder
+          update_versions = false
```
ferventcoder: This is a placeholder for @Iristyle to add - for the `--latest` \ `--update` parameter for updating the resources. Alternatively we could just delete the yaml file and let it update everything but having something we can pass in will also work.

dsc_resource_release_tags.yml
```diff
@@ -0,0 +1,48 @@
...
+xSQLServer: 1.7.0.0-PSGallery
+xStorage: 2.6.0.0-PSGallery
+xTimeZone: 1.5.0.0-PSGallery
+xWebAdministration: 1.12.0.0-PSGallery
+xWebDeploy: 1.2.0.0-PSGallery
+xWindowsEventForwarding: 1.0.0.0-PSGallery
+xWindowsRestore: 1.0.0-PSGallery
+xWindowsUpdate: 2.5.0.0-PSGallery
+xWinEventLog: 1.1.0.0-PSGallery
+xWordPress: 1.1.0.0-PSGallery
```
ferventcoder: @puppetlabs/windows may not be a bad idea for someone to go through and verify these are all correct. (and file tickets with Microsoft if they are not).

ferventcoder:
Alright @Iristyle - handoff. There is only adding in the `--latest`\`--update` switch left for you to implement and for folks to review.

build/dsc.rake
```diff
@@ -67,16 +71,54 @@ eod
...
+      resource_tags = {}
+      resource_tags = YAML::load_file("#{dsc_resources_file}") if File.exists? dsc_resources_file
+
+      puts "Getting latest release tags for DSC resources..."
+      Dir["#{dsc_resources_path_tmp}/xDSCResources/*"].each do |dsc_resource_path|
+        dsc_resource_name = Pathname.new(dsc_resource_path).basename
+        FileUtils.cd(dsc_resource_path) do
+          # --date-order probably doesn't matter
+          tags_raw = %x{ git log --tags --pretty=format:'%D' --simplify-by-decoration --date-order }
+          # if this starts to result in errors, we should explore pushing this out to a file
```
ferventcoder: s/file/out to a method where we can clean up the tags that may have prerelease versions in them.

build/dsc.rake
ferventcoder: This is complete!

ferventcoder:
I did not generate the resources from this to allow the review of this. I think that could/would be done in a followup PR closely following the acceptance of this.

build/dsc.rake
ferventcoder: I didn't remove it as it doesn't hurt anything being there.

build/dsc.rake
```diff
@@ -67,16 +72,54 @@ eod
...
+      Dir["#{dsc_resources_path_tmp}/xDSCResources/*"].each do |dsc_resource_path|
+        dsc_resource_name = Pathname.new(dsc_resource_path).basename
+        FileUtils.cd(dsc_resource_path) do
+          # --date-order probably doesn't matter
+          tags_raw = %x{ git log --tags --pretty=format:'%D' --simplify-by-decoration --date-order }
+          # If the conversion of string to version starts to result in errors,
+          # we should explore pushing this out out to a method where we can
+          # clean up the tags that may have prerelease versions in them
+          # similar to what was done in the Chocolatey module
+          versions = tags_raw.scan(/(\S+)\-PSGallery/).map { | ver | Gem::Version.new(ver[0]) }
```
ferventcoder: Scan is not my favorite item, but it works.

ferventcoder:
Once comments are looked over and addressed here, we can close this in favor of #219.

ferventcoder:
Unless I should move my comments over.

ferventcoder:
Superseded by #219

PR #219, opened 2016-07-19
(MODULES-3228) Track DSC Resources Based On Release Versions

glennsarti:
### Use PSGallery Release Tags
Each submodule tracks on a commit SHA1, but that SHA1 doesn't
necessarily correspond or represent an actual release of that
submodule (aka DSC Resource) to the PowerShell Gallery. However,
those resources are tagged when they are released to the gallery
with *-PSGallery, where * represents the version going up to the
gallery. This allows someone to find the currently released
version of a module/DSC Resource.

Instead of trusting the tags that are available for a resource in
the list of submodules, which tends to be completely inaccurate if
you are looking for the latest release, gather all PSGallery tags
and find the latest version. Then check that version out for
building the resource.

This code is similar to what James proposed in de5a10eac0.

Paired with James Pogran <james.pogran at puppet.com>
Paired with Glenn Sarti <glenn.sarti at puppet.com>

### Track release versions
Generate and track versions being used in a file. When running, if not
updating, use the versions being tracked to ensure compatibility.

The versions can be PSGallery tags (preferred), branches or commit SHAs.

If a tracked version is not found when building, use/record the latest
version of that released resource. This allows the ability to add new
dsc resources automatically.

To update to the latest version of each resource, pass `true` as an
argument for `update_versions` to the import rake task or delete the
yaml file tracking the release versions.

### Initial Tracked Version list

For this PR, the tracked version list `dsc_resource_release_tags.yml` is set to the
commit SHAs for latest released DSC module (DSC_REF=e6974e1).  A later Pull Request
will update the tracked versions with the latest ones from PS Gallery tags.

### Outstanding items:

 * [x] Add a yaml to track versions
 * [x] Use the yaml by default
 * [x] `--latest` to attempt to update the versions

#### Open questions 

* [x] ~~For new items that are added or dropped, how should that be handled?~~ If the item is not already in the YAML, it will be added with the latest tag and can be adjusted manually. For items that are removed, well they will just need to be removed from the yaml eventually. That or one of the ways to update to the latest versions is to delete the yaml and let it rebuild. This will also have the effect of removing items that are no longer available. Then source control diffs will show you what has changed.


glennsarti:
This supersedes PR #218

glennsarti:
Modifications from PR 218:
- Added ability to track on commit or branch name
- Modified tracked version list to mirror the current DSC module.

build/dsc.rake
```diff
@@ -73,25 +77,52 @@ eod
...
+          else
+            puts "Using tag for #{dsc_resource_name} is #{tracked_version}. Switching to that tag."
+            checkout_version = tracked_version
+          end
+
+          # If the checkout_version is not a standard PSGallery tag, a git fetch
+          # is required before a git checkout e.g. for commits or non-default branch names
+          if !(checkout_version =~ /-PSGallery/)
+            puts "#{checkout_version} is not a PSGallery tag. Fetching from git remote"
+            sh "git fetch"
```
ferventcoder: The submodule update doesn't handle this?

ferventcoder: Thinking this could have been added as an additional commit...

glennsarti: No because I would then be depending on DSC_REF variable still.  This way you can use any commit or branch in history.

It seemed such a small change, and it was still in the spirit of the previous commits (using a specific point in time of the repo) already so I didn't see it was required.  Can pull it out to a separate one if required.

ferventcoder: No, it's fine.

build/dsc.rake
```diff
@@ -67,16 +72,61 @@ eod
...
+          versions = tags_raw.scan(/(\S+)\-PSGallery/).map { | ver | Gem::Version.new(ver[0]) }
+          if versions.empty?
+            raise "#{dsc_resource_name} does not have any '*-PSGallery' tags. Appears it has not been released yet. Tags found #{tags_raw.to_s}"
+          end
+
+          latest_version = versions.max.to_s + "-PSGallery"
+          tracked_version = resource_tags["#{dsc_resource_name}"]
+
+          update_version = update_versions
+          update_version = true if tracked_version.nil?
```
MosesMendoza: Not sure if the team generally goes with the tertiary operator or not but would this be easier to grok as 
`update_version = tracked_version.nil? ? true : update_versions`
or even
`update_version = tracked_version.nil? ? true : args[:update_version]`
?

build/dsc.rake
```diff
@@ -67,16 +72,61 @@ eod
...
+
+          if update_version
+            puts "Latest/available tag for #{dsc_resource_name} is #{latest_version}. Switching to that tag."
+            checkout_version = latest_version
+          else
+            puts "Using tag for #{dsc_resource_name} is #{tracked_version}. Switching to that tag."
```
MosesMendoza: Minor grammatical suggestion - Maybe something like `"Using tag #{tracked_version} for #{dsc_resource_name}. Switching to that tag."`

build/dsc.rake
ferventcoder: Does it work in 1.9.3? If so I love it. 👍

build/dsc.rake
ferventcoder: We also need to update this to remove the word tag as it could just be a sha. @glennsarti are you updating these?

build/dsc.rake
MosesMendoza: Yes, 1.9.3 compatible

glennsarti: Updated.

build/dsc.rake
glennsarti: Updated to `Using the latest/available reference of #{latest_version} for #{dsc_resource_name}.` and `Using the specific......`

ferventcoder:
👍

PR #222, opened 2016-07-26
(MODULES-3640) Update modulesync 30fc4ab

MosesMendoza:
See MODULES-3640 for more context. tldr; json_pure as of 2.0.2 requires ruby 2 - because Puppet tests against many different versions of ruby, including 1.9.3, we need to ensure we can still get json_pure. The original json_pure dependency is expressed by puppetlabs/puppet (unversioned).


ferventcoder:
AppVeyor, you so cray cray

ferventcoder:
👍

PR #223, opened 2016-07-27
(MODULES-3491) Describe version strategy in README

glennsarti:
This commit documents the version strategy used to determine a version number
when releasing the module.  This commit also removes trailing whitespace.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ In this version, the following DSC Resources are already built and ready for use
 
  - PowerShell 5, which is included in [Windows Management Framework 5.0][wmf-5.0].
  - [Windows 2003 is not supported](#known-issues).
- 
+
 ## Setup
 
 ~~~
@@ -406,7 +406,7 @@ Error: Try 'puppet help module install' for usage
   ~~~
 
   For Puppet 4.2.2+ (and 3.8.2) we've decreased the possibility of the issue occurring based on the fixes in [PUP-4854](https://tickets.puppet.com/browse/PUP-4854). A complete fix is plannd in a future version of Puppet that incorporates [PUP-4866](https://tickets.puppet.com/browse/PUP-4866).
-  
+
   If you are affected by this issue:
   - Use the `--module_working_dir` parameter to set a different temporary directory which has a smaller length, for example;
     `puppet module install puppetlabs-dsc --module_working_dir C:\Windows\Temp`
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -447,6 +447,14 @@ For more information, see our [module contribution guide.](https://docs.puppet.c
 * The Puppet types are built from the source code of each DSC Resources MOF schema files. If you want to build the types, read the [Quick-start and Building](README_BUILD.md#quick-start).
 * If you want the build Puppet types for your own custom DSC Resources, read [Build Custom DSC Resource Types](README_BUILD.md#build-custom-dsc-resource-types).
 
+### Version Strategy
+
+This module generally follows [Semantic Versioning](http://semver.org/) for choosing an appropriate release version number with the following exception;
+
+* Minor e.g. from version 2.0.0 to 2.1.0
+
+A minor change may include [rebuilding the DSC resource types](README_BUILD.md).  Puppet wants to keep pace with the released DSC Resources from the PowerShell team repository, but this engenders risk as Puppet adopts third party code. Normally this would mean making major version bumps, but since this is anticipated to be frequent that would be too much churn. Instead the minor version will indicate when types are regenerated.
```
MosesMendoza: Question on the last sentence, `Instead the minor version will indicate when types are regenerated.`

I read that "will" as indicating that we will always/only change the minor version when regenerating types. Is that accurate? If not perhaps something like "Instead a minor version change may indicate that types have been regenerated."

glennsarti: Fair point.  Will update.

README.md
```diff
@@ -447,6 +447,14 @@ For more information, see our [module contribution guide.](https://docs.puppet.c
 * The Puppet types are built from the source code of each DSC Resources MOF schema files. If you want to build the types, read the [Quick-start and Building](README_BUILD.md#quick-start).
 * If you want the build Puppet types for your own custom DSC Resources, read [Build Custom DSC Resource Types](README_BUILD.md#build-custom-dsc-resource-types).
 
+### Version Strategy
+
+This module generally follows [Semantic Versioning](http://semver.org/) for choosing an appropriate release version number with the following exception;
```
jtappa: colon here, instead of a semicolon

README.md
```diff
@@ -447,6 +447,14 @@ For more information, see our [module contribution guide.](https://docs.puppet.c
...
+
+* Minor e.g. from version 2.0.0 to 2.1.0
```
jtappa: We prefer "for example" instead of "e.g." for localization purposes.

README.md
jtappa: * I agree with Moses, we try to avoid "will".
* There's a double space after the first sentence that needs removing.
* Lowercase the R in `DSC Resources`.

Thanks!

MosesMendoza:
👍

PR #225, opened 2016-07-28
(MODULES-3627) Allow PowerShell dependency up to v3

ferventcoder:
The currently released version of PowerShell at v2 is now
considered compatible with DSC, so allow installing a newer
version of PowerShell when using the DSC module.

### DIFF
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -1,10 +1,7 @@
 fixtures:
   forge_modules:
     stdlib: "puppetlabs/stdlib"
-    powershell:
-      repo: "puppetlabs/powershell"
-      # pin to version 1.0.6 to match metadata version string reqs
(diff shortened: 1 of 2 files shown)


ferventcoder:
@puppetlabs/windows please review.

Iristyle:
👍 for increased compatibility as a stop-gap solution.

Would like to see us just remove the dep altogether though.

Iristyle:
@ferventcoder do you want to update `.fixtures.yml` to match this as well?

ferventcoder:
@Iristyle done. This is ready again.

ferventcoder:
> Would like to see us just remove the dep altogether though.

Agreed. I think we have a ticket to address that. That removal comes with removing the `lcm_config`, so it will be a major version boundary.

PR #228, opened 2016-08-08
(maint) Add test artifacts to PMT Ignore

glennsarti:
Previously some artifacts created from testing, for example Beaker, the DSC
module are accidentally added to the package during a `puppet module build`.
This commit adds log, junit and tmp to the Puppet Module Tool ignore file.


MosesMendoza:
👍

PR #232, opened 2016-08-24
(MODULES-3764) Fix WaitForAll blacklist test

ferventcoder:
With https://github.com/puppetlabs/puppet/commit/01db61e24 in Puppet 4.6.x+ we
see that with https://github.com/puppetlabs/puppet/blob/5fcc9b3f803b62e84287e1976a5f1a2eb6311f27/lib/puppet/pops/evaluator/runtime3_resource_support.rb#L31
the message has changed and we need to adjust the matcher to account for both
the old and new way of this message occurring.

As an example, the message for `dsc_waitforsome` when not found as
a valid type was changed from `Invalid resource type dsc_waitforsome`
to `Unknown resource type: 'dsc_waitforsome'`, causing the matching
to fail.

### DIFF
--- a/tests/acceptance/tests/basic_dsc_resources/waitforall/negative/waitforall_blacklist.rb
+++ b/tests/acceptance/tests/basic_dsc_resources/waitforall/negative/waitforall_blacklist.rb
@@ -18,7 +18,7 @@ dsc_manifest_template_path = File.join(local_files_root_path, 'basic_dsc_resourc
 dsc_manifest = ERB.new(File.read(dsc_manifest_template_path), 0, '>').result(binding)
 
 # Verify
-error_msg = /Error:.*Invalid resource type dsc_waitforall/
+error_msg = /Error:.*resource.*dsc_waitforall/
 
 # Tests
 agents.each do |agent|
--- a/tests/acceptance/tests/basic_dsc_resources/waitforany/negative/waitforany_blacklist.rb
+++ b/tests/acceptance/tests/basic_dsc_resources/waitforany/negative/waitforany_blacklist.rb
@@ -18,7 +18,7 @@ dsc_manifest_template_path = File.join(local_files_root_path, 'basic_dsc_resourc
 dsc_manifest = ERB.new(File.read(dsc_manifest_template_path), 0, '>').result(binding)
 
 # Verify
-error_msg = /Error:.*Invalid resource type dsc_waitforany/
+error_msg = /Error:.*resource.*dsc_waitforany/
 
 # Tests
 agents.each do |agent|
--- a/tests/acceptance/tests/basic_dsc_resources/waitforsome/negative/waitforsome_blacklist.rb
+++ b/tests/acceptance/tests/basic_dsc_resources/waitforsome/negative/waitforsome_blacklist.rb
@@ -18,7 +18,7 @@ dsc_manifest_template_path = File.join(local_files_root_path, 'basic_dsc_resourc
 dsc_manifest = ERB.new(File.read(dsc_manifest_template_path), 0, '>').result(binding)
 
 # Verify
-error_msg = /Error:.*Invalid resource type dsc_waitforsome/
+error_msg = /Error:.*resource.*dsc_waitforsome/
 
 # Tests
 agents.each do |agent|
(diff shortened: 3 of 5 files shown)


tests/acceptance/tests/basic_dsc_resources/waitforall/negative/waitforall_blacklist.rb
```diff
@@ -18,7 +18,7 @@
 dsc_manifest = ERB.new(File.read(dsc_manifest_template_path), 0, '>').result(binding)
 
 # Verify
-error_msg = /Error:.*Invalid resource type dsc_waitforall/
+error_msg = /Error:.*resource.*dsc_waitforall/
```
Iristyle: Do we want to make sure there's whitespace around `resource`?

```
error_msg = /Error:.*\s+resource\s+.*\s+dsc_waitforall/
```

Iristyle:
Do we want to mention what the message changed from / to in commit msg?

From: `Invalid resource type dsc_waitforsome`
To: `Unknown resource type: 'dsc_waitforsome'`

tests/acceptance/tests/basic_dsc_resources/waitforall/negative/waitforall_blacklist.rb
ferventcoder: We could, I thought about it. I decided against it in case someone comes along and adds `'Invalid resource'` or something to that effect in a similar manner to what was done with `type` to `'type'`

ferventcoder:
@Iristyle done on the commit message. I'm hesitant to ensure space just due to things like `Invalid resource:` and `invalid 'resource'`

ferventcoder:
I think the tests capture the gist of what we expect, an error with the resource name in it and something like "resource" in the message. 👍

tests/acceptance/tests/basic_dsc_resources/waitforall/negative/waitforall_blacklist.rb
MosesMendoza: @ferventcoder curious why you removed the literal 'type' from between resource and <resource name>, e.g., why not `error_msg = /Error:.*resource type.*dsc_waitforall/`

ferventcoder: Looking for the wording that is necessary to indicate that this error was both related to a resource and that resource's name was `dsc_waitforall`. The change to me seems the most resilient to future changes. 

(String matching)--

ferventcoder: I don't see "resource" being part of too many words that would fall into the context of a false positive here.

ferventcoder: Or to put the argument another way, for the same reason "Invalid" in the statement "Invalid resource type" no longer works. Because it is another word that could change, introduce punctuation, or contain line feeds between. Granted, it's very unlikely to change. 

Of course we'd hoped that "Invalid" didn't change and it did, although I think it makes more sense in context that it is now "Unknown" instead. I would just rather not see tests fail when someone decides to change wording.

MosesMendoza: Makes sense. I was just wondering if there would be some other error that bubbles up that is masked by the regex, but I think that's an outside chance given what we're talking about here. 👍

MosesMendoza:
👍

Iristyle:
We also need to change the other 2 tests:

* https://github.com/puppetlabs/puppetlabs-dsc/blob/master/tests/acceptance/tests/basic_dsc_resources/waitforany/negative/waitforany_blacklist.rb
* https://github.com/puppetlabs/puppetlabs-dsc/blob/master/tests/acceptance/tests/basic_dsc_resources/waitforsome/negative/waitforsome_blacklist.rb

Once we do that, then I'm 👍

ferventcoder:
Added those, plus the two additional blacklisted items `xChrome` and `xFirefox`.

PR #236, opened 2016-09-21
(FM-5671) PsDscRunAsCredential Support

jpogran:
This adds the PsDscRunAsCredential parameter to all Puppet types
generated from all the DSC Resources.

WMF5 DSC added the PsDscRunAsCredential to every DSC Resource, which
allows the the DSC engine to execute a DSC Resource under a user context
instead of the SYSTEM context the DSC engine normally runs under. This
parameter is dynamically added to all DSC Resources at runtime and not
part of the schema.mof files we rely on to generate our types. This
means we have to 'hardcode' it in our type generation.

The File DSC Resource schema was modified to remove PsDscRunAsCredential
as it was erroneously added when the file was first added to the repo.
This file is only for us when we build types, so it does not affect any
vendor DSC Resources

### DIFF
--- a/build/dsc/templates/dsc_type.rb.erb
+++ b/build/dsc/templates/dsc_type.rb.erb
@@ -95,6 +95,22 @@ Puppet::Type.newtype(:dsc_<%= resource.friendlyname.downcase %>) do
     defaultto { :<%= resource.default_ensure_value %> }
   end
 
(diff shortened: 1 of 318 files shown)


jpogran:
@puppetlabs/windows please review

ferventcoder:
Reviewing.

PR #242, opened 2016-10-04
(MODULES-3935) Update to DSC Resource September Reskit Release

jpogran:
This updates all DSC Resources to the Septemeber DSC Resource Kit
release

This should wait until #253 gets merged first


Iristyle:
#247 should be finished / merged prior to this.

PR #243, opened 2016-10-05
Do not use (MODULES-3683) Add Supported DSC Resources to build system

jpogran:
This adds the supported DSC Resources to the type building system by
moving both the xDscResources and the dscresources folder to a new
staging folder called thedescresources. This is a hacky fix, but avoids
having to modify the MOF parsing and building classes to support reading
more than one folder.

This also changes the custom DSC resource detection we did previously.
It relied on the directory name being xDscResources, since we have moved
things this now looks for thedscresources instead.


Iristyle:
#247 should be finished / merged prior to this.

PR #244, opened 2016-10-05
(FM-5662) Reduce the DSC type tests

jpogran:
This reduces the amount of boilerplate type tests generated by building
the types from the dsc resources. The original intention was to cover
all the types, but as we continue to keep pace with the released DSC
Resources the number of tests increase exponentionally without adding
any extra value. This change selects specific DSC Resources in order to
keep the safety of the unit tests without the exceedingly long runtime
penalty of the redundant tests.

This should be merged before #242 and #243 to reduce the time to test for those


Iristyle:
Based on AppVeyor results, its unclear to me that this is actually working as desired.

jpogran:
I didn't rebuild and remove the unwanted tests, wanted to solicit feedback first. Will put up another commit with the rebuild

Iristyle:
Also - I assume this is just a stop-gap before merging the validation between regular MOF / embedded instance MOF data and stripping out the unnecessary tests from the templates?

ferventcoder:
> Also - I assume this is just a stop-gap before merging the validation between regular MOF / embedded instance MOF data and stripping out the unnecessary tests from the templates?

What does this mean?

Iristyle:
@ferventcoder - I look at https://tickets.puppetlabs.com/browse/MODULES-2760 as the real solution to the spec problem - this PR is a nice fix for reducing the test count, but addresses the symptom rather than the underlying problem.

I'll spike something real quick to show you what that would look like.

ferventcoder:
> I'll spike something real quick to show you what that would look like.

If you want, that's fine @Iristyle. At the very least this is definitely a great start, shaving nearly 45 minutes off of the AppVeyor build time, so this is good to move forward on for now.

Iristyle:
The validation refactoring is in #246

PR #245, opened 2016-10-07
(maint) Fix rake dsc:build to update if needed

Iristyle:
- Currently, rake dsc:build skips an import of the upstream DSC
   resource from github.com/PowerShell/DscResources.  However, this
   is problematic when the local yml file has been updated, but the
   built types don't match locally - resulting in unpredictable
   behavior.

 - Instead, always call the import handler, which will look at the
   local git cloned copy of the upstream resources, and will make sure
   to fetch from upstream repos when the desired tags are not present.

 - Rework the file cleanup for the cloned copy of the source, so that
   original files are not deleted - instead filter them when copying.
   Don't trash the cloned source at the end of the import.


Iristyle:
@puppetlabs/windows this should fix the rake task issues I was encountering.

Iristyle:
@jpogran I've noticed that a bunch of the links in the table are wrong as
well in types.md

PR #249, opened 2016-10-07
WIP - MOF parsing

Iristyle:
@jpogran I had at one point started patching the MOF parsing process to be smarter - but I never finished it off.

This is the WIP point-in-time code that I had started (but obviously some things have changed since then).

I also had a repo that I was working off of that I had put together as a test at https://github.com/Iristyle/dsc-parser-tests/blob/master/parse.rb that might contain some useful tidbits.

### DIFF
--- a/build/dsc/mof.rb
+++ b/build/dsc/mof.rb
@@ -4,8 +4,6 @@ module Dsc
     attr_accessor :dsc_invalid_resources
 
     def initialize(options)
-
-      @import_folder                = options[:import_folder]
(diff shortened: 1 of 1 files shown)


Iristyle:
There's some more detail in a gist at https://gist.github.com/Iristyle/cbd1f57ce53de9964202

david22swan:
@Iristyle 
Closing this PR as the work enclosed has been largely replaced by puppetlabs-dsc_lite.

PR #254, opened 2016-10-26
(MODULES-3683) Fix maxpath issue with SharePointDsc and OfficeOfflineDsc

jpogran:
Implementing MODULES-3683 introduced maxpath errors with the OfficeOnlineServerDsc and SharePointDsc as they have a child folder
structure that is different than all the other DSC Resources we have
encountered so far.

OfficeOfflineServerDsc and SharePointDsc both have a three level deep
folder structure until you actually get to the DSC Resource. This is
different than all the other DSC Resources in
https://github.com/PowerShell/DscResources. If you look at their packaging
script, they account for it by only going to the actual DSC Resource:
https://github.com/PowerShell/SharePointDsc/blob/dev/.appveyor/appveyor.psm1#L59
when they make their nuget package for the PSGallery.

This fixes this by inspecting the path of each file and looking for
{ModuleName}/Modules/{ModuleName} and removing the extra folder.


jpogran:
This was tested by building the module and then installing the module using puppet:

```
PS C:\puppetlabs\modules\dsc> bundle exec puppet module build
DL is deprecated, please use Fiddle
Notice: Building C:/puppetlabs/modules/dsc for release
Module built: C:/puppetlabs/modules/dsc/pkg/puppetlabs-dsc-1.1.0.tar.gz
PS C:\puppetlabs\modules\dsc> bundle exec puppet module install C:\puppetlabs\modules\dsc\pkg\puppetlabs-dsc-1.1.0.tar.g
z
DL is deprecated, please use Fiddle
Notice: Preparing to install into C:/ProgramData/PuppetLabs/code/modules ...
Notice: Downloading from https://forgeapi.puppetlabs.com ...
Notice: Installing -- do not interrupt ...
C:/ProgramData/PuppetLabs/code/modules
└─┬ puppetlabs-dsc (v1.1.0)
  ├── puppetlabs-powershell (v2.0.3)
  ├── puppetlabs-reboot (v1.2.1)
  └── puppetlabs-stdlib (v4.13.1)
```

ferventcoder:
LGTM!

PR #248, opened 2016-10-07
(MODULES-4014) Fail if the PowerShell version is not >= 5

Iristyle:
- Users have been presented with confusing messages around type
  suitability that don't clearly indicate that this module requires
  PowerShell 5 or higher.
  
  Fail hard when the minimum PS version is not met.

PR #261, opened 2016-11-29
(MODULES-3977) DSC - WMF 5.1 fix

jpogran:
The dsc module uses the PowerShell Manager code from the powershell
module. The powershell module was recently updated to work on WMF 5.1
and uses named pipes. This PR adds this fix to the dsc module.


glennsarti:
@jpogran Looks like we'll need to update the spec tests as well.  Appveyor timed out.

jpogran:
@glennsarti I think we hit a temporary problem with appveyor. puppet 3.0 and ruby 1.9.3 took 1 min, but puppet 3.0 and ruby 2.0 took over 1 hour? re-kicking this

lib/puppet/provider/base_dsc/powershell.rb
```diff
@@ -21,23 +22,27 @@
...
-Unfortunately this library prevents the DSC module from using a shared
-PowerShell process to dramatically improve the performance of resource
-application.
-
-To enable these improvements, it is suggested to upgrade to any x64 version of
-Puppet (including 3.x), or to a Puppet version newer than 3.x.
-    UPGRADE
-  end
-
+  POWERSHELL_UPGRADE_MSG = <<-UPGRADE
```
Iristyle: Looks like you have superseded the proposed work in #248 / MODULES-4014 - which is great!

However, I'm not sure that this message makes the most sense for the DSC module.  DSC module requires PowerShell 5 since we're not using the `DscConfiguration` style cmdlets.

I think we can get away with something simpler here that is more specific to the DSC module (and I think as a result you can close out that other PR and the MODULES-4014 ticket at the same time)

lib/puppet_x/puppetlabs/compatible_powershell_version.rb
```diff
@@ -0,0 +1,51 @@
+require File.join(File.dirname(__FILE__), 'powershell_version')
+
+module PuppetX
+  module PuppetLabs
+    module Dsc
+      class CompatiblePowerShellVersion
+        def self.compatible_version?
```
Iristyle: As we were discussing in HipChat ... probably want to mod the implementation of `self.compatible_version?` to something as simple as:

```ruby
def self.compatible_version?
  powershell_version = PuppetX::PuppetLabs::Dsc::PowerShellVersion.version

  return false if powershell_version.nil?

  Gem::Version.new(powershell_version) >= Gem::Version.new(5)
end
```

lib/puppet/provider/base_dsc/powershell.rb
jpogran: This will be done in the MODULES-4014 ticket as it likely needs a test and QA review, as we are restricting on version where we didn't before

lib/puppet_x/puppetlabs/compatible_powershell_version.rb
glennsarti: Will be addressed in https://github.com/puppetlabs/puppetlabs-dsc/pull/292

PR #265, opened 2016-12-19
(FM-5972) Update to next modulesync_configs [dedaf10]

DavidS:


ferventcoder:
@puppetlabs/modules can this be closed? I think we already updated this in a different PR?

Iristyle:
Correct, this is no longer relevant.  Closing.

PR #282, opened 2017-02-20
(MODULES-4422) PS Module Manifest Parser bug

jpogran:
This fixes the regex that parses the PS module mainfest file (psd1) so that it can handle both single and double quoted strings


glennsarti:
Do we need some tests around this?

nealpeters86:
Tested with 3 custom modules;
The fix is working for the cChoco module, also building 2 other custom modules is still working as expected! Thanks @jpogran!

jpogran:
@nealpeters86 thanks for the verification. I am still going to put some kind of spec test for this like @glennsarti says, as it's obviously a blind spot

glennsarti:
@jpogran CI is green

glennsarti:
Given a customer has verified the fix, and my commit was to fix up CI tests due to Gem loading, while normally I would not merge my own commits, in this case this seems ok.

PR #283, opened 2017-02-22
(MODULES-4430) Fix repo case change

jpogran:
This fixes the case change in the official DSCResources folder name in the PowerShell/DSCResources repo. It also moves the folder names to a variable, and uses the variable in all the places it was specified before, so that it is now a change once fix

### DIFF
--- a/build/dsc.rake
+++ b/build/dsc.rake
@@ -70,13 +70,15 @@ eod
 
         sh cmd
 
+        community_dsc_resources_root = "#{dsc_resources_path_tmp}/xDscResources"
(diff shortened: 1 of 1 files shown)


glennsarti:
Do you have a link to the commit in the DSC repo for the case change?

jpogran:
Putting it here, will also put it in the commit: https://github.com/PowerShell/DscResources/commit/7ab39f90b735c2cb62b3a23c6e2c39747869ec96

jpogran:
Commit updated. Ready for review again @glennsarti

PR #290, opened 2017-03-22
[MODULES-4528] modulesync e25ca9 - Add locales folder and config.yaml and update default nodeset

wilson208:


wilson208:
@glennsarti have updated the message now

PR #292, opened 2017-04-03
(Modules-4014) Restrict to Powershell v5

jpogran:
This restricts the dsc module from running on PowerShell versions less than 5

~~**NOTE** - PR #261 needs to be merged first~~

### DIFF
--- a/lib/puppet/provider/base_dsc/powershell.rb
+++ b/lib/puppet/provider/base_dsc/powershell.rb
@@ -22,26 +22,32 @@ Puppet::Type.type(:base_dsc).provide(:powershell) do
 Applies DSC Resources by generating a configuration file and applying it.
 EOT
 
-  POWERSHELL_UPGRADE_MSG = <<-UPGRADE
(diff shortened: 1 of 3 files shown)


lib/puppet/provider/base_dsc/powershell.rb
```diff
@@ -78,6 +84,9 @@ def exists?
     Puppet.debug "PowerShell Version: #{version}"
     script_content = ps_script_content('test')
     Puppet.debug "\n" + script_content
+    
+    fail self.class.POWERSHELL_UPGRADE_MSG if !PuppetX::Dsc::PowerShellManager.compatible_version_of_powershell?
```
glennsarti: Minor nit:  Ruby usually uses `unless` instead of `if not <blah>`  Not sure it improves readability though

```
fail self.class.POWERSHELL_UPGRADE_MSG unless PuppetX::Dsc::PowerShellManager.compatible_version_of_powershell?
```

PR #294, opened 2017-04-07
(MODULES-4667) Add facade rake tasks for test tiering

glennsarti:
Previously, the Jenkins based CI testing called beaker directly instead of using
a standard rake task.  This commit adds two facade style rake tasks called
`acceptance_tests` and `reference_tests` as per the standard Puppet test tiering
practices.  This commit then just calls the equivalent beaker command.  In the
future these take tasks would call Rototiller etc. appropriately.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -20,3 +20,33 @@ PuppetLint.configuration.send("disable_class_inherits_from_params_class")
 PuppetLint.configuration.send("disable_80chars")
 
 task :default => [:spec, :lint]
+
+# These rake tasks exist to pre-empt adding test tiering and rototiller the DSC Module
+# Note that these are very Puppet Jenkins CI specific
+desc "Rake facade for running acceptance tests"
+task :acceptance_tests do
+  # Environment variables normally used for Puppetlabs Spec Helper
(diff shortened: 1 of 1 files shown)


Rakefile
```diff
@@ -20,3 +20,31 @@ PuppetLint.configuration.send("disable_class_inherits_from_params_class")
...
+# Note that these are very Puppet Jenkins CI specific
+desc "Rake facade for running acceptance tests"
+task :acceptance_tests do
+  # Environment variables normally used for Puppetlabs Spec Helper
+  hostsfile = ENV['BEAKER_setfile']
+  keyfile   = ENV['BEAKER_keyfile']
+
+  system("beaker --color --debug --preserve-hosts never --hosts #{hostsfile} --keyfile #{keyfile} --load-path tests/lib --pre-suite tests/integration/pre-suite --tests tests/integration/tests")
+
+  if cmd_status != 0
```
eputnam: should this be `$?.exitstatus`? does cmd_status also work?
https://ruby-doc.org/core-2.2.0/Kernel.html#method-i-system

glennsarti: Fixed

PR #296, opened 2017-04-19
(maint) Ensure no PendingFileRenameOperations prior to reboot test

glennsarti:
Previously the VM template used during acceptance testing had a pending
file rename operation which caused reboot based tests to fail.  This commit adds
a step to remove the offending registry key prior to testing.  This workaround
will not be required once https://tickets.puppetlabs.com/browse/IMAGES-539 is
resolved.


glennsarti:
Adhoc passed with success.  Can you merge @jpogran ?

PR #306, opened 2017-07-25
(MODULES-5228) Fix PowerShellManager template path

jpogran:
This moves the init_ps1.ps1 PowerShellManager template to a subdirectory so that
there are no conflicts with other modules that use `lib/puppet_x/templates` path.


glennsarti:
Relates to https://github.com/puppetlabs/puppetlabs-powershell/pull/193

glennsarti:
Merging even though Appveyor failed.  This is due to Puppet 4.2.3 testing, but that version is not supported.  Only cell not run is Pup5 and 64 bit Ruby, but 32bit succeed and given the changes this is ok.

PR #309, opened 2017-08-09
(MODULES-5534) Update to July DSC Resource Kit Release

jpogran:
This commit updates to the current July DSC Resource Kit Release


Iristyle:
Moved ticket to [MODULES-5534](https://tickets.puppetlabs.com/browse/MODULES-5534) as JIRA integration doesn't work for PRs (and probably other things?) I'm surprised there's no git subject validation for the tix like what's used in the Puppet project itself, but modules land may just be out of sync with what other projects / teams do.

Do we have a formal process for looking at the updated DSC content, or is it simply to the point of being a quick look at the files to make sure they make sense? (Ideally this could be automated in AppVeyor / Travis)

jpogran:
No formal process other than making sure it runs through adhoc first. I think this is moot for this particular PR since we'll be doing another PR right away for the August release

PR #311, opened 2017-08-23
(MODULES-3409) Add deprecation warning for powershell and stdlib modules

jpogran:
This adds a warning that we will remove the dsc::lcm_config manifest, which removes puppetlabs-powershell module and puppetlabs-stdlib.

Configuring the LCM was only necessary in the preview version of WMF 5.0, and is not needed for puppet to function with dsc in the RTM version which we moved to in MODULES-2770.

### DIFF
--- a/README.md
+++ b/README.md
@@ -338,6 +338,8 @@ All [Puppet metaparameters](https://docs.puppet.com/references/latest/metaparame
 
 ### Optionally Configure the DSC LCM RefreshMode
 
+*WARNING* The dsc::lcm_config class will be removed in the v1.5 release of this module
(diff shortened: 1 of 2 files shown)


jpogran:
I think travis had a hiccup, running the same thing locally as travis with same ruby/puppet versions doesn't error. This is ready for review.

*Note* We shouldn't be too concerned about messaging here as it's a low-use feature that is being removed in the next release

Iristyle:
I trigged Travis multiple times and it still failed... though I agree there shouldn't be a reason it's failing.

Merging

PR #312, opened 2017-08-23
(MODULES-2980) Move vendored resources test

jpogran:
This commit moves the vendored resources test to acceptance to ensure that a problem is found sooner


Iristyle:
The original PR that introduced this is at #58 

I don't see any reason that test needed to be via `puppet agent` rather than `puppet apply`. +1 to not involving the master.

PR #317, opened 2017-08-29
(maint) Fix minor linting error in lcm_config

glennsarti:
The Puppet Lint configuration for the log format was not working as intended and
was masking the Lint errors.  This commit removs the configuration item as it
is not needed.

This commit updates the lcm_config to use single quotes as per puppet lint style
guidelines.


Iristyle:
This was a result of merging #311 which was failing Travis at the time with an obscure message.  I should have realized that we needed to give it further review given it failed Travis multiple times with an obscure `The command "bundle exec rake release_checks" exited with 1.` without any further detail.

PR #319, opened 2017-08-30
Fix dsc_xpowershellexecutionpolicy with non localsystem account 

GerbenWelter:
If 'puppet agent' is started as an user with local admin privileges, Puppet keeps applying dsc_xpowershellexecutionpolicy even though the policy is already set correctly.

Checking and setting the PowerShell Execution Policy within the scope of LocalMachine fixes this issue.

### DIFF
--- a/lib/puppet_x/dsc_resources/xPowerShellExecutionPolicy/DSCResources/MSFT_xPowerShellExecutionPolicy/MSFT_xPowerShellExecutionPolicy.psm1
+++ b/lib/puppet_x/dsc_resources/xPowerShellExecutionPolicy/DSCResources/MSFT_xPowerShellExecutionPolicy/MSFT_xPowerShellExecutionPolicy.psm1
@@ -35,7 +35,7 @@ function Set-TargetResource
         Try
         {
             Write-Verbose "Setting the execution policy of PowerShell."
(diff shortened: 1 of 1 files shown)


jpogran:
Thanks @GerbenWelter for the PR, but we can't modify the DSC Resources directly in this repo as they get pulled from the source git repos for each DSC Resource when we build the module, and your code will get overwritten.

You will have to make a PR against the [xPowerShellExecutionPolicy](https://github.com/PowerShell/xPowerShellExecutionPolicy) repo in order to get this fix merged. Since this fix is specific to the user environment of each server, I would suggest making the scope a new parameter for the DSC Resource in your PR. However you approach it, once they merge your PR and make a release for xPowerShellExecutionPolicy, we can add it to the puppet dsc module.

GerbenWelter:
Thank you @jpogran for the info. I'll make the PR to the correct repository.

I get what you are saying about adding a variable for the scope but from a Puppet and maybe from a DSC point of view it might not be needed because you are managing the server the code is running on. Not a remote server.

GerbenWelter:
I see it's already a known issue: PowerShell/xPowerShellExecutionPolicy#8

I'll go from there.

PR #321, opened 2017-09-06
(maint) Fix PowerShell manifest error checking

jpogran:
This commit changes the psd1 manifest parsing code to check for nil
instead of empty on the hash of attributes from the manifest.
Previously if the hash did not have a 'ModuleVersion' because of a
failed parse, the code looked to see if the hash was empty, causing an
exception that does not contain the DSC Resource that failed parsing.
Using nil instead keeps the full formed error message

### DIFF
--- a/build/dsc/psmodule.rb
+++ b/build/dsc/psmodule.rb
@@ -10,7 +10,7 @@ module Dsc
     end
 
     def version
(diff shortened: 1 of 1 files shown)


Iristyle:
Closing in favor of #322 which includes this already and has comments about a minor fix to this commit.

PR #322, opened 2017-09-06
(MODULES-5590) Fix PowerShell Manifest Parsing semicolons

jpogran:
This PR fixes the regex that parses the DSC Resource module manifest
files to correctly parse the module version.

A PowerShell psd1 manifest file follows the syntax for a PowerShell
hash. It allows newlines as well as semi-colons to deliminate each item
in the hash. It is not normal syntax to use semi-colons inside a hash
that is using newlines already, but it is still valid so the regex is
adjusted to accept it.

### DIFF
--- a/build/dsc/psmodule.rb
+++ b/build/dsc/psmodule.rb
@@ -10,7 +10,7 @@ module Dsc
     end
 
     def version
-      raise "ModuleVersion not found for module #{@name} / #{@module_manifest_path}" if attributes['ModuleVersion'].empty?
+      raise "ModuleVersion not found for module #{@name} / #{@module_manifest_path}" if !attributes.key?('ModuleVersion') || attributes['ModuleVersion'].empty?
       attributes['ModuleVersion']
     end
 
--- a/build/dsc/psmodule.rb
+++ b/build/dsc/psmodule.rb
@@ -18,7 +18,7 @@ module Dsc
       begin
         unless @attributes
           attrs = {}
-          regex = /^(.*) *= *['"](.*)['"] *$/
+          regex = /^(.*) *= *['"](.*)['"] *(;)? *$/
           File.open(@module_manifest_path, 'r') do |psd1|
             content = File.read(psd1)
             utf8_encoded_content = utf8_encode_content(content)


build/dsc/psmodule.rb
```diff
@@ -10,15 +10,15 @@ def initialize(module_name, module_manifest_path)
     end
 
     def version
-      raise "ModuleVersion not found for module #{@name} / #{@module_manifest_path}" if attributes['ModuleVersion'].empty?
+      raise "ModuleVersion not found for module #{@name} / #{@module_manifest_path}" if attributes['ModuleVersion'].nil?
```
Iristyle: This changes the original intent a bit.

There are a few things that could be tested here:

* 1: `attributes` is `nil` - not necessary to do because the `attributes` method later in the class raises if the `psd1` can't be parsed
* 2: `attributes` doesn't have a `ModuleVersion` at all (i.e. the `PSD1` omitted it)
* 3: `attributes` has a `ModuleVersion`, but it's an empty string (i.e. the `PSD1` defined it like `ModuleVersion = ''`)
* 4: `attributes` has a `ModuleVersion` and the value is `nil` - not necessary to do based on how the `attributes` method does regex captures

We were previously checking for 3, but here it's changed to 2, but I think it should be both 2 and 3, i.e.

```
[9] pry(main)> foo = {'ModuleVersion' => ''}
=> {"ModuleVersion"=>""}
[10] pry(main)> foo.key?('ModuleVersion')
=> true
[11] pry(main)> foo['ModuleVersion'].nil?
=> false
[12] pry(main)> foo['ModuleVersion'].empty?
=> true
```

Also, `.key?` is usually used to check for hash existence because values can be set `nil`.

So we'll need something like:

```ruby
if !attributes.key?('ModuleVersion') || attributes['ModuleVersion'].empty?
```

Where error raised in first 2 instances, but 3rd is OK

```
[1] pry(main)> foo = {}; !foo.key?('ModuleVersion') || foo['ModuleVersion'].empty?
=> true
[2] pry(main)> foo = { 'ModuleVersion' => '' }; !foo.key?('ModuleVersion') || foo['ModuleVersion'].empty?
=> true
[3] pry(main)> foo = { 'ModuleVersion' => 'val' }; !foo.key?('ModuleVersion') || foo['ModuleVersion'].empty?
=> false
```

jpogran: `attributes`can be nil, because the quoting slips by the regex resulting in an empty hash, not an exception.

```
From: /Users/jpogran/src/puppetlabs/puppetlabs-dsc/build/dsc/psmodule.rb @ line 16 Dsc::Psmodule#version:

    12: def version
    13:   if attributes['ModuleVersion'].nil?
    14:     require 'pry';binding.pry
    15:   end
 => 16:   raise "ModuleVersion not found for module #{@name} / #{@module_manifest_path}" if attributes['ModuleVersion'].empty?
    17:   attributes['ModuleVersion']
    18: end

[1] pry(#<Dsc::Psmodule>)> attributes
=> {"Description"=>"The XenDesktop7 DSC resources can automate the deployment and configuration of Citrix XenDesktop 7.x. These DSC resources are provided AS IS, and are not supported through any means."}
[2] pry(#<Dsc::Psmodule>)>
```

This allows things to continue until `empty?` is called, but that throws an exception when the key is not there, thereby squashing the 'real' problem...that the manifest was parsed incorrectly.

``rake aborted!
NoMethodError: undefined method `empty?' for nil:NilClass
/Users/jpogran/src/puppetlabs/puppetlabs-dsc/build/dsc/psmodule.rb:13:in `version'
(erb):81:in `block (2 levels) in build_dsc_types'
/Users/jpogran/src/puppetlabs/puppetlabs-dsc/build/dsc/manager.rb:187:in `block (2 levels) in build_dsc_types'
/Users/jpogran/src/puppetlabs/puppetlabs-dsc/build/dsc/manager.rb:186:in `open'
/Users/jpogran/src/puppetlabs/puppetlabs-dsc/build/dsc/manager.rb:186:in `block in build_dsc_types'
/Users/jpogran/src/puppetlabs/puppetlabs-dsc/build/dsc/manager.rb:178:in `each'
/Users/jpogran/src/puppetlabs/puppetlabs-dsc/build/dsc/manager.rb:178:in `build_dsc_types'
/Users/jpogran/src/puppetlabs/puppetlabs-dsc/build/dsc.rake:228:in `block (3 levels) in <top (required)>'
/Users/jpogran/src/puppetlabs/puppetlabs-dsc/.bundle/gems/ruby/2.3.0/gems/rake-12.3.0/exe/rake:27:in `<top (required)>'
/Users/jpogran/.rbenv/versions/2.3.5/bin/bundle:22:in `load'
/Users/jpogran/.rbenv/versions/2.3.5/bin/bundle:22:in `<main>'
Tasks: TOP => dsc:types:build
(See full trace by running task with --trace)
```

The proper fix is to have the attributes method actually test for the attributes that we use when building types instead of relying on a simplistic error handling in the version method, but this was a quick fix at the time to get a customer working.

**Comment by jpogran** [comment]
Rebased and updated branch.

Amended maint commit to use key and empty

File: build/dsc/psmodule.rb [hunk 2]
```diff
@@ -18,7 +18,7 @@ def attributes
       begin
         unless @attributes
           attrs = {}
-          regex = /^(.*) *= *['"](.*)['"] *$/
+          regex = /^(.*) *= *['"](.*)['";] *$/
```
Iristyle: This regex now allows `;` to terminate a string like this:

```
[2] pry(main)> "foo = 'bar;" =~ /^(.*) *= *['"](.*)['";] *$/
=> 0
```

Even though this regex doesn't do the balanced quote matching properly, we probably want something that captures an optional `;` in the right spot like `/^(.*) *= *['"](.*)['"] *(;)? *$/` (normally I'd use `\s` for whitespace, but since this regex is already using a literal `<space>` I left that bit as-is)


```
[3] pry(main)> "foo = 'bar;" =~ /^(.*) *= *['"](.*)['"] *(;)? *$/
=> nil

[4] pry(main)> "foo = 'bar';" =~ /^(.*) *= *['"](.*)['"] *(;)? *$/
=> 0

[5] pry(main)> "foo = 'bar'" =~ /^(.*) *= *['"](.*)['"] *(;)? *$/
=> 0

[6] pry(main)> "foo = \"bar\"" =~ /^(.*) *= *['"](.*)['"] *(;)? *$/
=> 0
```

jpogran: Since you already went through the extra work I'll add it, but I don't think it's worth our time to be defensive against improper syntax. The format you suggest wouldn't have passed Test-ModuleManifest or the gallery. I don't want to be in the business of re-implementing the PowerShell tokenizer, as using regex is on the path to [evil](https://stackoverflow.com/questions/1732348/regex-match-open-tags-except-xhtml-self-contained-tags)

jpogran:
This is ready for review again @Iristyle

build/dsc/psmodule.rb
Iristyle: Balanced HTML tags is a much much more difficult task than balancing quotes - don't fear the regex! ;0

I should have also pointed out your mod only matched because of the `(.*)` in the middle (which really should be `(\w*)`) greedily includes the trailing `'`, which means it returns the wrong captures and then sets the wrong value at line 28:

```
### original suggestion - notice capture 2 has a trailing ' after bar
[4] pry(main)> matches = "foo = 'bar';".match(/^(.*) *= *['"](.*)['";] *$/)
=> #<MatchData "foo = 'bar';" 1:"foo " 2:"bar'">

### new version - capture 2 doesn't have the trailing ' after bar
[5] pry(main)> matches = "foo = 'bar';".match(/^(.*) *= *['"](.*)['"] *(;)? *$/)
=> #<MatchData "foo = 'bar';" 1:"foo " 2:"bar" 3:";">
```

So this needed to be changed for that reason

PR #324, opened 2017-10-16
(MODULES-5743) Support Sensitive in MSFT_Credential

Iristyle:
- Previously a password could not be specified in a MSFT_Credential
   based attribute of a given resource, due to strict validation that
   was only allowing strings.

   Sensitive was initially added in Puppet 4.6.0 and fully fleshed out
   by Puppet 4.8.0. This module currently supports a minimum of
   Puppet 4.7.0 and therefore users should be expected to use
   Sensitive for password fields (example updated in README).

 - Added new tests for type helpers demonstrating the ability for
   MSFT_Credential to allow Sensitive.

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,6 +175,8 @@ dsc_example_resource { 'examplefoo':
 
 DSC uses `MSFT_Credential` objects to pass credentials to DSC Resources. Supply a hash to any `credential` parameter, and Puppet handles creating the `credential` object for you.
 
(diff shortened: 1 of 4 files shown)


michaeltlombardi:
This has passed adhoc and manual testing performed.

glennsarti:
@Iristyle You'll need a maint commit to remove Puppet 4.2.3 testing (sync.yml and appveyor.yml)

Iristyle:
#324 can be merged, then this can be rekicked

Iristyle:
Removal of openssl upgrade is in #332

PR #329, opened 2017-11-28
(MODULES-4748) Prevent zombie procs from timeout

RandomNoun7:
Prior to this commit if a timeout occurs while waiting for the
connection to a named pipe then a zombie process is left on the
system as the code to close or kill it is never called. This commit
ensures that such zombie processes are removed before the error is
raised.

This commit also adds an RSPEC test to validate that if we try to
create a manager with an unreasonably low timeout value, that the
manager will not have time to open the pipes server and it will
fail to create the manager and further verify that no zombie
processes are left running.

Paired-with: Bill Hurt <william.hurt@puppet.com>

### DIFF
--- a/lib/puppet_x/puppetlabs/powershell_manager.rb
+++ b/lib/puppet_x/puppetlabs/powershell_manager.rb
@@ -9,14 +9,14 @@ module PuppetX
     class PowerShellManager
       @@instances = {}
 
-      def self.instance(cmd, debug = false)
+      def self.instance(cmd, debug = false, pipe_timeout = 30)
         key = cmd + debug.to_s
         manager = @@instances[key]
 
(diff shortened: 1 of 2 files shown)


Iristyle:
Whoa... @RandomNoun7 paired with himself on this one @michaeltlombardi !

RandomNoun7:
Poo. I kinda thought it would be a little more obvious that this was just a copy of @michaeltlombardi 's original PR. Guess not.

michaeltlombardi:
@RandomNoun7 ➕ ➕ for self-pairing!

This code looks good.... suspiciously good....

michaeltlombardi:
Ad-Hoc passed, merging.

PR #331, opened 2017-12-19
(maint) Remove AppVeyor 4.2.3 Puppet testing

Iristyle:
- Manually removed the extra modulesync addition for testing
   Puppet 4.2.3 which is no longer necessary, and updated
   AppVeyor accordingly.

### DIFF
--- a/.sync.yml
+++ b/.sync.yml
@@ -29,9 +29,6 @@ spec/spec_helper.rb:
   - 'vendor/'
 appveyor.yml:
   appveyor_bundle_install: "bundle install --jobs 4 --retry 2 --without system_tests build"
-  matrix_extras:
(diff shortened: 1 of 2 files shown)


Iristyle:
There is no explanation for the 4.2.3 addition in https://github.com/puppetlabs/puppetlabs-dsc/commit/b5e5d895b2635c8565f3c1a9367bdce1e5b5a495 other than "sync travis.yml" despite there being quite a few more files touched.  The ticket at https://tickets.puppetlabs.com/browse/MODULES-4097 has a little more detail, but not a ton.

Since the current module metadata.json requires Puppet 4.7.0+ at https://github.com/puppetlabs/puppetlabs-dsc/blob/master/metadata.json#L32 it should be safe to bump minimum Puppet version now.

This is necessary for #324

PR #333, opened 2017-12-19
(maint) modulesync cd884db Remove AppVeyor OpenSSL update on Ruby 2.4

michaeltlombardi:


Iristyle:
I don't think we actually need this?  Confused about the diff since all of that should already be in `master`

PR #335, opened 2018-01-02
(MAINT) Fix travis config

michaeltlombardi:
This commit updates the Travis config to remove the erroneous
`before_install` setting which is causing CI to fail. This
should be captured in the near future in a modsync update.


michaeltlombardi:
aw yeah

PR #337, opened 2018-01-08
(RE-9941) Migrate from stickler to artifactory

sarameisburger:
update tests to point to new internal gem host


michaeltlombardi:
Ninja'd by @glennsarti!

PR #353, opened 2018-03-27
(IMAGES-795) 2008r2 Failing PowerShell Tests

RandomNoun7:
The newest image for Windows 2008r2 is exhibiting different behavior
for loading .NET assemblies than we have seen in the past. The
System.Core assembly is not loaded into the PowerShell host process by
default any longer. This causes the init.ps1 script in the module to
to fail to start a Pipe Server to receive PowerShell scripts to be
executed. This change always attempts to load the System.Core
assembly prior to creating the pipe server object. Explicity loading the
assembly fixes the issue and the newest image becomes usable again. This
fix is also require to fix any moduels that depend on the PowerShell
exec provider working properly.

### DIFF
--- a/lib/puppet_x/templates/dsc/init_ps.ps1
+++ b/lib/puppet_x/templates/dsc/init_ps.ps1
@@ -736,6 +736,8 @@ function Start-PipeServer
     $Encoding
   )
 
+  Add-Type -AssemblyName "System.Core, Version=3.5.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089"
+
   # this does not require versioning in the payload as client / server are tightly coupled
   $server = New-Object System.IO.Pipes.NamedPipeServerStream($CommandChannelPipeName,
     [System.IO.Pipes.PipeDirection]::InOut)


RandomNoun7:
This PR is a copy of the same one merged for PowerShell module - [puppetlabs/puppetlabs-powershell#214](https://github.com/puppetlabs/puppetlabs-powershell/pull/214)

Iristyle:
We shouldn't be using `IMAGES` tickets for modules.

lib/puppet_x/templates/dsc/init_ps.ps1
```diff
@@ -736,6 +736,8 @@ function Start-PipeServer
     $Encoding
   )
 
+  Add-Type -AssemblyName "System.Core, Version=3.5.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089"
```
Iristyle: What about for systems that don't have the `3.5.0.0` version of `System.Core`?

Iristyle: More commentary on https://github.com/puppetlabs/puppetlabs-powershell/pull/214#discussion_r177478151

PR #356, opened 2018-04-17
(MODULES-6839) Convert Service and User tests.

RandomNoun7:
This change converts the Service and User tests to beaker-rspec.


spec/acceptance/basic_dsc_resources/service/service_spec.rb
```diff
@@ -0,0 +1,128 @@
+require 'spec_helper_acceptance'
+
+describe 'Service tests' do
+
+  context 'MODULES-2336 - C68749 - Attempt to Apply DSC Service Resource with Incompatible "StartupType" and "State"' do
+
+    # ERB Manifest
```
ThoughtCrhyme: These are no longer and ERB templates

spec/acceptance/basic_dsc_resources/service/service_spec.rb
```diff
@@ -0,0 +1,128 @@
+require 'spec_helper_acceptance'
+
+describe 'Service tests' do
+
+  context 'MODULES-2336 - C68749 - Attempt to Apply DSC Service Resource with Incompatible "StartupType" and "State"' do
```
ThoughtCrhyme: The "C68749" is a TestRail id, which we're not going to be using in the future so we can get rid of that. Also the JIRA ticket is kind of distracting in output, so maybe move that to a comment.

spec/acceptance/basic_dsc_resources/service/service_spec.rb
```diff
@@ -0,0 +1,128 @@
...
+
+    windows_agents.each do | agent |
+      it 'Should apply a manifest with no errors.' do
+        on(agent, puppet('apply'), :stdin => dsc_manifest, :acceptable_exit_codes => [0,2]) do |result|
+          assert_no_match(/Error:/, result.stderr, 'Unexpected error was detected!')
+        end
+      end
+
+      it 'Should verify results' do
+        expect_failure('Expected to fail because of MODULES-2512') do
```
ThoughtCrhyme: MODULES-2512 is resolved as fixed, so this either should work, or should fail for a different reason from MODULES-2512

spec/acceptance/basic_dsc_resources/user/user_spec.rb
```diff
@@ -0,0 +1,219 @@
...
+          password,
+          dsc_type,
+          dsc_module,
+          :Password,
+          :Ensure   => dsc_props[:dsc_ensure],
+          :UserName => username
+        )
+      end
+
+      it 'Should perform teardown' do
```
ThoughtCrhyme: typically teardown is done in an "after(:all)" block

RandomNoun7: I agree and attempted to do, but the function definition for `assert_dsc_resource` actually contains an `assert` if you follow the definition down far enough. I can't put an example in `before` or `after` blocks. The assert means this is actually a test. Is a refactor for this in scope for this ticket? Because if it is, it also contains a redundant implementation of a function to execute powershell that I wouldn't mind getting rid of.

spec/acceptance/basic_dsc_resources/service/service_spec.rb
RandomNoun7: Checking.

RandomNoun7: The fix for MODULES-2512 was a [documentation fix only](https://github.com/puppetlabs/puppetlabs-dsc/commit/5e8873957f1ff528ee0975bd1a401f0f822e39e8#diff-04c6e90faac2675aa89e2176d2eec7d8R348). The test still passes because the manifest constructed and passed into `assert_dsc_resource` does not follow the guidelines recommended in the docs change. In effect, this is a canary test that will simply tell us if MS ever fixes this bug.

ThoughtCrhyme: Word, makes sense.

spec/acceptance/basic_dsc_resources/user/user_spec.rb
ThoughtCrhyme: Oh yeah, hmmm, I don't think we want to refactor that `assert_dsc_resource` method, but should scope that for future work.

RandomNoun7:
@ThoughtCrhyme I don't know why github thinks I dismissed your review, but if you want to go ahead and approve this beast I'd be much obliged.

spec/spec_helper.rb
```diff
@@ -3,7 +3,7 @@
 require 'fileutils'
 
 # for debugging
-require 'pry' if Bundler.rubygems.find_name('pry').any?
+#require 'pry' if Bundler.rubygems.find_name('pry').any?
```
RandomNoun7: @ThoughtCrhyme Well, hold on a sec, I need to figure out how this got in here.

PR #361, opened 2018-04-24
(WIP)(maint) Make build tasks work on windows

jpogran:
This commit enables building this module on windows platforms by
removing the charlock_holmes dependency and updating all file read/write
methods to use binary methods.

The charlock_holmes gem is only used to parse psd1 manifest files, so
the removal is constrained to one location. The gem is replaced with
inline encoding enforcement to UTF8. We don't need to worry about all
the variances that charlock_holmes worries about, we only read the psd1
file to get the version number. This is also less important as we move
to PSGallery imports and deprecate git imports.

All relevant file methods were updated to use binary reads and writes to
avoid CRLF differences when writing and reading files on windows. This
setting will 'do the right thing' regardless on windows or on linux/mac.

Note, this is dependent on PR https://github.com/puppetlabs/mof/pull/5
which updates the MOF parser to handle absolute paths on windows.

### DIFF
--- a/build/dsc.rb
+++ b/build/dsc.rb
@@ -11,7 +11,6 @@ require 'logger'
 if Bundler.rubygems.find_name('mof').any?
   require 'mof'
   require 'cim'
-  require 'charlock_holmes/string'
 end
 
 # Debug
--- a/build/dsc/property.rb
+++ b/build/dsc/property.rb
@@ -21,6 +21,7 @@ module Dsc
       begin
         if @description.nil? && @cim_feature.description
           content = @cim_feature.description
+          content.force_encoding(Encoding::UTF_8)
           content.gsub!(/\xe2\x80\x9c/u, "'") # usage of single quotes as this string will be injected in a double bracket "#{content}"
           content.gsub!(/\xe2\x80\x9d/u, "'") # usage of single quotes as this string will be injected in a double bracket "#{content}"
           content.gsub!(/\xe2\x80\x98/u, "'")
--- a/build/dsc/psmodule.rb
+++ b/build/dsc/psmodule.rb
@@ -3,6 +3,13 @@ module Dsc
 
     attr_accessor :name
 
+    ENCODING = Encoding::UTF_8
+    ENCODING_OPTIONS = {
+      invalid: :replace,
+      undef:   :replace,
+      replace: ''
+    }.freeze
+
     def initialize(module_name, module_manifest_path)
       @name = module_name
       @module_manifest_path = module_manifest_path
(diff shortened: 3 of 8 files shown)


Iristyle:
FWIW - `charlock_holmes` can work on Windows, but if we can remove it from the process like this, even better.

build/dsc.rb
```diff
@@ -11,7 +11,6 @@
 if Bundler.rubygems.find_name('mof').any?
   require 'mof'
   require 'cim'
-  require 'charlock_holmes/string'
```
Iristyle: FYI, this is a monkey patch against Ruby `String` which reimplements `force_encoding` - so I would grep the code for that to find any usages just in case.

build/dsc/psmodule.rb
```diff
@@ -39,8 +49,11 @@ def attributes
     private
 
     def utf8_encode_content(content)
-      detection = CharlockHolmes::EncodingDetector.detect(content)
-      CharlockHolmes::Converter.convert content, detection[:encoding], 'UTF-8'
+      content.force_encoding(ENCODING)
+      unless content.valid_encoding?
+        content.encode!(ENCODING, ENCODING_OPTIONS)
```
Iristyle: This overwrites the strings encoding with UTF-8, then tries to re-encode it as UTF-8 if not all byte sequences are valid, substituting replacement characters for byte sequences it doesn't understand... `valid_encoding?` is not a well named method because it doesn't really tell you if that string is actually really UTF-8 or not (which is a much more complex problem that charlock_holmes solves by leveraging ICU).

This won't work at all if the string is originally UCS-2 / UTF-16 or its something like ISO-8859-1, 1252, etc. Do we know if there's enforcement for PSD1 to be UTF-8 anywhere?

Note that the `File.open` above will default to using `Encoding.default_external` which is generally UTF-8 on Linux, but will be current codepage on Windows, which can lead to a world of hurt if local codepage is something like 932 and the file contains legit UTF-8 characters ;0

That said, we can solve this problem... I would call `File.read` with `utf-8|bom` on a first read attempt and maybe just look for `ModuleVersion` as you've alluded to, instead of reading all the attributes. If we can't find that we could try to fallback to reloading the file as `utf-16le|bom` and if we get no matches, abort reading the PSD1. We at least know all the characters involved in that string are ASCII and can make some concessions on how valid our file reading methodology is.

I can go into more detail if it's not clear what I'm suggesting.

build/dsc/property.rb
```diff
@@ -21,6 +21,7 @@ def description
       begin
         if @description.nil? && @cim_feature.description
           content = @cim_feature.description
+          content.force_encoding(Encoding::UTF_8)
```
Iristyle: I'm a little concerned about the use of `force_encoding` here as it wipes out the strings encoding. Was this a result of loading files as `BINARY`? FWIW, it's better to explicitly specify the encoding when opening the file for reading in the first place (like `'BOM|UTF-8'` - see https://ruby-doc.org/core-2.0.0/IO.html#method-c-new-label-IO+Encoding)

glennsarti:
@jamesp Needs a rebase

PR #363, opened 2018-04-26
(MODULES-7063) Fix case sensitive psd1 search

jpogran:
This commit fixes the code that finds the psd1 manifest file for a given
DSC Resource.

A psd1 manifest is assumed to be the same name as the DSC Resource. However
on case sensitive OS, the file name could be a different case than the DSC
Resource name, i.e SqlServerDsc vs SQLServerDsc.ps1, which would cause
File.exists? to fail to find the file even though it exists.

Dir.glob is used here to find any psd1 manifest in the module_dir folder.
This is a safe assumption because by convention there can only be one
manifest file for a DSC Resource because it follows PowerShell Module
manifest conventions. To address cases where there are more than one
psd1 file in a DSC Resource directory, we filter on module name,
lowercased.

If both methods have failed to find a psd1 manifest,
then we truly do not have a manifest, or just haven't accounted for yet
another variance

### DIFF
--- a/build/dsc/resource.rb
+++ b/build/dsc/resource.rb
@@ -186,8 +186,25 @@ module Dsc
         end
 
         module_dir = path_array[0..(path_array.count - (index + 2))].join('/')
-        module_manifest_path = "#{module_dir}/#{module_name}.psd1"
-        raise "module manifest #{module_manifest_path} not found" unless File.exists?(module_manifest_path)
+
+        # A psd1 manifest is assumed to be the same name as the DSC Resource
+        # However on case sensitive OS, the file name could be a different
+        # case than the DSC Resource name, i.e SqlServerDsc vs SQLServerDsc.ps1,
+        # which would cause File.exists? to fail to find the file even though
+        # it exists.
+        #
+        # Dir.glob is used here to find any psd1 manifest in the module_dir
+        # folder. This is a safe assumption because by convention there can
+        # only be one manifest file for a DSC Resource because it follows
+        # PowerShell Module manifest conventions. However there could be cases
+        # where there are other psd1 files for different purposes, so we
+        # filter for ones that match the module_name exactly, lowercased
+        module_manifest_path = Dir.glob("#{module_dir}/*.psd1")
+                                  .select{ |fn| (File.basename(fn, ".psd1")).casecmp(module_name) == 0 }
+                                  .first
+        if module_manifest_path.nil?
+          raise "module manifest #{module_dir}/#{module_name}.psd1 not found"
+        end
 
         @ps_module = Dsc::Psmodule.new(module_name, module_manifest_path)
       end
--- a/build/dsc/psmodule.rb
+++ b/build/dsc/psmodule.rb
@@ -2,6 +2,7 @@ module Dsc
   class Psmodule
 
     attr_accessor :name
+    attr_reader :module_manifest_path
 
     def initialize(module_name, module_manifest_path)
       @name = module_name
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -46,3 +46,21 @@ end
 Puppet::Util::Log.level = :warning
 Puppet::Util::Log.newdestination(:console)
 
+# Mocking Objects
+class MockCIMClass
+  attr_accessor :name
+  attr_accessor :qualifiers
+
+  def initialize(name)
+    @name = name
+    @qualifiers = {}
+  end
+end
+
+class MockCIMQualifier
+  attr_accessor :value
+
+  def initialize(value)
+    @value = value
+  end
+end
--- a/spec/unit/build/dsc/resource_spec.rb
+++ b/spec/unit/build/dsc/resource_spec.rb
@@ -1,0 +1,63 @@
+require 'spec_helper'
+require File.join(File.dirname(__FILE__), '../../../../build/dsc/resource')
+require File.join(File.dirname(__FILE__), '../../../../build/dsc/psmodule')
+
+fixture_path = File.expand_path(File.join(__FILE__, '..', '..', '..', '..', 'fixtures'))
+
+def create_mock_cim_class(name, friendlyname)
+  obj = MockCIMClass.new(name)
+  obj.qualifiers['Friendlyname'] = MockCIMQualifier.new(friendlyname)
+
+  obj
+end
+
+describe 'Dsc::Resource' do
+  let(:subject) { Dsc::Resource.new(mof_class, mof_path) }
+
+  describe '#ps_module' do
+    context 'with a default style DSC Resource module' do
+      let(:mof_class) { create_mock_cim_class('default', 'defaultfriendly') }
+      let(:mof_path) { File.join(fixture_path, 'dsc_modules', 'default', 'DSCResources')}
+
+      it 'should create a PS Module object' do
+        result = subject.ps_module
+
+        expect(result.name).to eq('default')
+        expect(result.module_manifest_path).to match(/\/default.psd1$/)
+      end
+    end
+
+    context 'with a PSD1 file which has different casing than the DSC Resource module ' do
+      let(:mof_class) { create_mock_cim_class('mixedCase', 'mixedCasefriendly') }
+      let(:mof_path) { File.join(fixture_path, 'dsc_modules', 'mixedCase', 'DSCResources')}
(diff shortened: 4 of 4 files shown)


glennsarti:
The PR makes sense.  I wonder if we should just use the glob method always?

jpogran:
> or just haven't accounted for yet another variance

So, I found a case where the convention I mentioned isn't followed:

```
[master ≡] ~\src\third\DscResources> ls -rec | ? extension -eq '.psd1' |
  group -property { [IO.Path]::GetDirectoryName("$($_.FullName)") } |
  ? count -gt 1 |
  fl

Name   : C:\Users\jpogran\src\third\DscResources\xDscResources\xActiveDirectory
Count  : 3
Group  : {HADCConfiguration.psd1, ParentChildConfig.psd1, xActiveDirectory.psd1}
Values : {C:\Users\jpogran\src\third\DscResources\xDscResources\xActiveDirectory}
```

So there could be cases where a DSC Resource folder can contains more than one psd1 file.

build/dsc/resource.rb
```diff
@@ -186,8 +186,31 @@ def ps_module
...
+          # which would cause File.exists? to fail to find the file even though
+          # it exists.
+          # Dir.glob is used here to find any psd1 manifest in the module_dir
+          # folder. This is a safe assumption because by convention there can
+          # only be one manifest file for a DSC Resource because it follows
+          # PowerShell Module manifest conventions. However there could be cases
+          # where there are other psd1 files for different purposes, so we
+          # filter for ones that match the module_name exactly, lowercased
+          glob_module_manifest_path = Dir.glob("#{module_dir}/*.psd1")
+                                         .select{ |fn| (File.basename(fn, ".psd1")).downcase == module_name.downcase }
```
Iristyle: I think `Glob` accepts a flag for ignoring case like `File::FNM_CASEFOLD` ... so I think you can reduce this down to something like this:

```ruby
Dir.glob("#{module_dir}/#{module_name}.psd1", File::FNM_CASEFOLD)
```

Iristyle: Given a folder like

```
Foo\fOo.psd1
```

`FNM_CASEFOLD` appears to do what you want:


Windows

```
[5] pry(main)> Dir.glob('Foo/Foo.psd1', File::FNM_CASEFOLD)
=> ["Foo/fOo.psd1"]
```

OSX

```
[1] pry(main)> Dir.glob('Foo/Foo.psd1', File::FNM_CASEFOLD)
=> ["Foo/fOo.psd1"]
```

glennsarti: I'm getting some odd behaviour with FNM_CASEFOLD

``` bash
[7] pry(#<Dsc::Resource>)> "#{module_dir}/#{module_name}.psd1"
=> "/mnt/c/Source/puppetlabs-dsc-james/spec/fixtures/dsc_modules/multiCase/multiCase.psd1"
[8] pry(#<Dsc::Resource>)> Dir.glob("#{module_dir}/#{module_name}.psd1", File::FNM_CASEFOLD)
=> []
[9] pry(#<Dsc::Resource>)> Dir.glob("#{module_dir}/*.psd1", File::FNM_CASEFOLD)
=> ["/mnt/c/Source/puppetlabs-dsc-james/spec/fixtures/dsc_modules/multiCase/Another.psd1",
 "/mnt/c/Source/puppetlabs-dsc-james/spec/fixtures/dsc_modules/multiCase/MuLTICase.psd1",
 "/mnt/c/Source/puppetlabs-dsc-james/spec/fixtures/dsc_modules/multiCase/ZzAnother.psd1"]
[10] pry(#<Dsc::Resource>)> Dir.glob("#{module_dir}/*#{module_name}.psd1", File::FNM_CASEFOLD)
=> ["/mnt/c/Source/puppetlabs-dsc-james/spec/fixtures/dsc_modules/multiCase/MuLTICase.psd1"]
[11] pry(#<Dsc::Resource>)>
```

It's like the case insensitivity only applied with a wildcard not a "direct" filename. Could be a WSL bug but doubt it.

glennsarti: So I tried this on Appveyor and Travis and the results are _interesting_
https://github.com/puppetlabs/puppetlabs-dsc/pull/364

Travis:
Ruby 2.1.9 was ok but Ruby 2.4.1 could use glob without the pattern

Appveyor:
Ruby 2.1.9 failed but RUby 2.4 was ok

TLDR; Using `File::FNM_CASEFOLD` without a pattern is too unreliable in this situation.

glennsarti: Using something similar to James' proposed method worked on all OSes and all ruby versions.

https://ci.appveyor.com/project/puppetlabs/puppetlabs-dsc/build/1.1.x.1173
https://travis-ci.org/puppetlabs/puppetlabs-dsc/builds/373345079

glennsarti:
You can cherry pick my tests from ;
https://github.com/puppetlabs/puppetlabs-dsc/pull/364/commits/cca6454fbe6f63acad27e078429b24e2118ba245 and https://github.com/puppetlabs/puppetlabs-dsc/pull/364/commits/f91267b2c2a14837e4d2633d9e5a8fc10265ce0b if you like.

build/dsc/resource.rb
```diff
@@ -186,8 +186,31 @@ def ps_module
...
+                                         .first
+          if !File.exists?(glob_module_manifest_path)
```
glennsarti: There is a subtle bug here.   If the glob doesn't pick anything it returns `nil` and then `File.exists?` raises an error that it can't do existance checks on nil.

Given that the glob already ensures that the file exists, I think you can safely just do `if glob_module_manifest_path.nil?`

build/dsc/resource.rb
Iristyle: Ah, WSL. But do we care about WSL?

For ha-has I tried it in the Ruby 2.5 docker container (based on Debian Stretch 9) and had a similar problem when wildcards weren't used.

```
root@7c6d029eb764:/# mkdir FOo
root@7c6d029eb764:/# touch FOo/fOo.psd1

root@7c6d029eb764:/# irb
irb(main):001:0> Dir.glob('foo/foo.psd1')
=> []
irb(main):002:0> Dir.glob('foo/foo.psd1', File::FNM_CASEFOLD)
=> []
irb(main):003:0> Dir.glob('foo/*.psd1', File::FNM_CASEFOLD)
=> []
irb(main):004:0> Dir.glob('*/*.psd1', File::FNM_CASEFOLD)
=> ["FOo/fOo.psd1"]
irb(main):005:0> RUBY_VERSION
=> "2.5.1"
```

Seems like a bug in the libc impl, because `FNM_CASEFOLD` is a flag for the `fnmatch` API in libc - http://man7.org/linux/man-pages/man3/fnmatch.3.html. Presumably WSL being Ubuntu based it uses a similar flavor / port of libc as Stretch 9 does, which would explain the consistency.

So even though none of us runs Linux, probably enough of a discrepancy to go the proposed route from @jpogran

Iristyle: Also we can use `casecmp` for case-insensitive compare in `.select{ |fn| (File.basename(fn, ".psd1")).downcase == module_name.downcase }` like `.select{ |fn| File.basename(fn, ".psd1").casecmp(module_name) == 0 }`

jpogran:
@glennsarti merged your commits and squashed/rebased and reworded commits. The appveyor failure is due to PowerShell_Manager timeouts

build/dsc/resource.rb
```diff
@@ -186,8 +186,25 @@ def ps_module
...
+        # it exists.
+        #
+        # Dir.glob is used here to find any psd1 manifest in the module_dir
+        # folder. This is a safe assumption because by convention there can
+        # only be one manifest file for a DSC Resource because it follows
+        # PowerShell Module manifest conventions. However there could be cases
+        # where there are other psd1 files for different purposes, so we
+        # filter for ones that match the module_name exactly, lowercased
+        module_manifest_path = Dir.glob("#{module_dir}/*.psd1")
+                                  .select{ |fn| (File.basename(fn, ".psd1")).casecmp(module_name) == 0 }
```
Iristyle: Since I updated my original comment, not sure if you saw that `.select{}.first` can be simplified to `.find` with the same predicate?

glennsarti: Didn't see that either.

PR #368, opened 2018-05-08
(MODULES-6742) Add Testmode Switcher

glennsarti:
This commit adds the testmode switcher gem to the Gemfile and acceptance testing
helper.  This is in preparation for modifying the tests to use the testmode
switcher methods to facilitate master and masterless testing.

---

Previously all of the DSC tests were agentless.  This commit modifies the tests
to use the testmode switcher methods (execute_manifest_xx), where appropriate.

* Some actions, such as creating and cleaning test fixtures must remain as
  agent (puppet apply) only
* This commit removes Beaker helper functions which are no longer called or
  required
* Adds two additional helper methods which bypass SSH and cygwin when creating
  files and folders.  This is due to IMAGES-825 causing test failures
* Fixes the 'File Resource with Valid Unicode "SourcePath" Specified' test in
  file_valid_spec.  This test should have been failing, however due to test
  state not being cleaned up prior, this test was accidentally passing.


glennsarti:
I have run this through adhoc succesfully.  The commit SHAs won't match up as I've rebased since then as I had to squash and reword commits

glennsarti:
@ThoughtCrhyme Update test names and found one you missed too.  Ready for re-review.

PR #371, opened 2018-05-15
(MODULES-5623) Handle DateTime values correctly

ananace:
DateTime values should be passed along as System.DateTime objects and not pure strings, this fixes issues with `starttime` property not being handled correctly.


jpogran:
Thanks for the contribution @ananace! Could you name a DSC Resource that uses datetime that this fixes so I can test this out?

ananace:
`dsc_xscheduledtask` is the one that was causing issues for us - and also the one mentioned in the ticket, so that's probably a good one for testing it.

lionce:
Hello @jpogran ,

Would you please check this PR? Should we merge it? It looks good to me! 

Thanks!

sheenaajay:
Hello @jpogran,
Would you please check this PR? Should we merge it? Thank you.

PR #374, opened 2018-08-13
(MODULES-7399) PDK convert the module

glennsarti:
This commit updates the module to be converted
to PDK format, and then converts the module to PDK format. There are no code or test
changes as part of this conversion. Note that rubocop testing is effectively
disabled and will be rectified in further pull requests.


glennsarti:
Running through adhoc

glennsarti:
Note that the adhoc git sha's are different as I noticed that some files were CRLF. I fixed them, but the commit changed.

PR #387, opened 2018-12-18
(MODULES-8172) Handle Sensitive for Credentials in Mismatched Environments

michaeltlombardi:


michaeltlombardi:
I made the decision (for now) to only modify munge for credentials because, as far as I can tell, it's the only place we supported sensitive values before - and, in my short research, everything I could think of that I'd want made sensitive is handled in DSC resources by shoving it into a credential.

I think this is the 95:5 use case and it will immediately resolve the customer issue for which this ticket was filed.
We will still need to rebuild all of the types, but given the scope and rabbit holes that presents, I thought it best to do in a separate PR.

PR #389, opened 2018-12-19
(MODULES-8172) Sensitive fixes across versions

Iristyle:
Builds on #387  

- Puppet 5 agents connected to Puppet 6 masters and Puppet 6 agents
   connected to Puppet 5 masters don't properly deserialize Sensitive
   values from catalogs.

   The type generation template has been updated with the inclusion
   of support to fix this.

   This commit includes the same fixes in the generated types.

 - Note: Due to issues with the build process, the rake tasks were not
   run over the existing content to produce these files.

   Instead, a regex search replace was performed using the following:

   search pattern:
   (?s)(validate_MSFT_Credential.*?end)

   replacement:
   $1\n    munge do |value|\n      PuppetX::Dsc::TypeHelpers.munge_sensitive_hash!(value)\n    end

### DIFF
--- a/lib/puppet/type/dsc_accountpolicy.rb
+++ b/lib/puppet/type/dsc_accountpolicy.rb
@@ -52,6 +52,9 @@ Puppet::Type.newtype(:dsc_accountpolicy) do
       end
       PuppetX::Dsc::TypeHelpers.validate_MSFT_Credential("Credential", value)
     end
+    munge do |value|
+      PuppetX::Dsc::TypeHelpers.munge_sensitive_hash!(value)
+    end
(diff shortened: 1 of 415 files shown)


Iristyle:
Since there are a lot of matches, we can just do a count, knowing that for every call to `validate_MSFT_Credential` should also have a call to `munge_sensitive_hash!`

Validations:

```
cat lib/puppet/type/* | grep -c validate_MSFT_Credential
720
```

Munges:

```
cat lib/puppet/type/* | grep -c munge_sensitive_hash!
720
```

It's a match!

PR #390, opened 2018-12-20
(MODULES-8395) PDK Sync  module to fix Travis failures.

RandomNoun7:


RandomNoun7:
There is a maint commit for the test fix mixed into this PR because the large number of new appveyor test cells makes it difficult to get a passing run in Appveyor without this fix in place. But it cannot be merged into master in it's own PR, because without the changes to Travis testing, Travis will fail and we can't merge. So it means two commit's in PR, but it needed to be done.

PR #391, opened 2018-12-20
(MAINT) Fix Powershell Manager Short Timeout Test

RandomNoun7:
This change fixes a race condition in the spec tests that causes
frequent transient errors and CI failures.


RandomNoun7:
Closing in favor of #390 since without the changes in that PR, this change cannot be merged.

PR #395, opened 2019-01-15
(MODULES-8389) Fix build process

michaeltlombardi:
Prior to this commit the build task would fail on any PowerShell
DSC module which did not have a tag in its repository which ended
in `-PSGallery`. Since our last release, this is true of multiple
modules included in the PowerShell DSC Resource module maintained
by Microsoft.

Even if a tracked version was specified the build still failed
because it checked for a tag regardless of whether the version
was tracked.

This commit updates the rake task for checkout to instead mark
the latest version as commit hash pinned for the submodule by
Microsoft if no PSGallery tags are found. This ensures that the
module code specified matches the code that Microsoft has pinned.

### DIFF
--- a/build/dsc.rake
+++ b/build/dsc.rake
@@ -97,10 +97,12 @@ namespace :dsc do
           # similar to what was done in the Chocolatey module
           versions = tags_raw.scan(/(\S+)\-PSGallery/).map { | ver | Gem::Version.new(ver[0]) }
           if versions.empty?
-            raise "#{dsc_resource_name} does not have any '*-PSGallery' tags. Appears it has not been released yet. Tags found #{tags_raw.to_s}"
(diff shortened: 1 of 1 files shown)


Iristyle:
That ended up being pretty straightforward!

Can you dump the build output as a comment in the PR here?

michaeltlombardi:
Yup, will do!

michaeltlombardi:
Failure from the pre-existing code:

https://gist.github.com/michaeltlombardi/67da150144954a5b635b8e4f5b498e53#file-unfixed-log-L489-L491

Log output from fixed build code:

https://gist.github.com/michaeltlombardi/67da150144954a5b635b8e4f5b498e53#file-fixed-log-L317-L322

PR #401, opened 2019-03-14
(MODULES-7689) fix module folder detection

jpogran:
This PR fixes:

- [x] Deals with HQRM module folder structure[1]
- [x] Moves all DSC Resource files back to structure that is psgallery structure
- [x] Removes orphaned/deleted files that were leftover from previous builds

It does increase build time and IO because it does several copies of all files instead of one, but I sacrifice IO speed and do more copies than technically needed to ensure the files stay in the structure they are supposed to be. There are a few special hardcoded cases because we are using the git repos and not the shipped psgallery versions. This can all be avoided when we move away from using the git repos.

Adhoc has passed.

To manually test:

- [x] Attempt to use any DSC Resource from 'ComputerManagementDsc', 'OfficeOnlineServerDsc',  or 'SharePointDsc.

[1] DSCResource folder being a subfolder to a Modules folder. For example: https://github.com/PowerShell/ComputerManagementDsc/tree/dev/Modules/ComputerManagementDsc/DSCResources


glennsarti:
@jpogran Is this still a WIP?

glennsarti:
Running through adhoc at https://jenkins-master-prod-1.delivery.puppetlabs.net/view/modules/view/windows/view/_adhoc/view/dsc/job/forge-windows_puppetlabs-dsc_init-manual-parameters_adhoc/45/

build/dsc.rake
```diff
@@ -61,13 +61,60 @@ namespace :dsc do
...
+        # essentially removing the `Modules` subfolder.
+        [ 'ComputerManagementDsc', 'OfficeOnlineServerDsc', 'SharePointDsc' ].each do | hq |
+          FileUtils.cp_r "#{dsc_resources_path_tmp}/DscResources/#{hq}", "#{default_dsc_module_path}/import/dsc_resources_copy"
+          FileUtils.cp_r "#{default_dsc_module_path}/import/dsc_resources_copy/#{hq}/Modules/#{hq}/.",
+            "#{default_dsc_module_path}/import/dsc_resources_copy/#{hq}"
+          FileUtils.rm_rf "#{default_dsc_module_path}/import/dsc_resources_copy/#{hq}/Modules/#{hq}"
+        end
+
+        # Because we're using the git repos, we have to get rid of all the files that don't get shipped
+        valid_files = m.find_valid_files("#{default_dsc_module_path}/import/dsc_resources_copy/**/*")
```
glennsarti: I know this technically isn't in the scope of this PR, but I _REALLY_ don't like single letter variables that span a lot of text

glennsarti:
Well..it's passed smoke and acceptance for 2019.0.x so it's looking good.

glennsarti:
Adhoc passed!

PR #405, opened 2019-04-11
(MODULES-8856) Redact debug

michaeltlombardi:
Prior to this commit the provider would output the full Powershell log to be executed during debug runs. This is to ensure that we can verify exactly what will be executed. However, this also meant that credentials and other secrets could leak into the debug output, and thus also into PuppetDB logs.

This commit modifies the provider and  ERB template to ensure that methods for formatting Puppet data to be PowerShell compliant are able to take a new parameter, `redact`.

Passing this parameter will cause the script builder to redact any sensitive data it is passed. This flag is forwarded along through a helper method. This implementation allows us to reuse the code for interpolating the correct values into the PowerShell script for the actual execution and for the debug messaging.

There is possibly a better implementation for this change.

This commit also includes updated spec tests to verify the behavior when choosing to redact sensitive values.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/) a
 
 ## [Unreleased]
 
+### Fixed
+- Ensure sensitive values are redacted in debug output ([MODULES-8856](https://tickets.puppetlabs.com/browse/MODULES-8856))
+
 ## [1.9.0] - 2019-04-02
 
 ### Added
--- a/lib/puppet/provider/base_dsc/powershell.rb
+++ b/lib/puppet/provider/base_dsc/powershell.rb
@@ -83,7 +83,7 @@ EOT
     version = Facter.value(:powershell_version)
     Puppet.debug "PowerShell Version: #{version}"
     script_content = ps_script_content('test')
-    Puppet.debug "\n" + script_content
+    Puppet.debug "\n" + self.class.redact_content(script_content)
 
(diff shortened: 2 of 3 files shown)


michaeltlombardi:
Changes passed in adhoc.

GerbenWelter:
This PR landed in version 1.9.1 and in my testing breaks when a DSC resource is used with the dsc_psrunascredential parameter.

When that happens a line in the DSC resource Puppet builds contains the following:

`psdscrunascredential = [PSCustomObject]@{'user' = 'some_user'; 'password' = 'very_secret' # PuppetSensitive} | new-pscredential`

But the '#' makes Powershell disregard the rest of the line and thus breaking the Puppet run.

The way I solved is by making the the added comment a multiline comment:

`psdscrunascredential = [PSCustomObject]@{'user' = 'some_user'; 'password' = 'very_secret' <# PuppetSensitive #>} | new-pscredential`

This also requires a change to the substitution Puppet does in the logging for the redacting but now the generated Powershell code is correct.

If interested I can make a new PR with the needed changes.

RandomNoun7:
Hello @GerbenWelter. Thanks very much for finding this and testing a fix. We always appreciate PR's for fixes like this, so please do submit one and we'll get it looked it!

RandomNoun7:
@GerbenWelter Please note that there is an almost identical fix in the dsc_lite repo (https://github.com/puppetlabs/puppetlabs-dsc_lite/pull/104). If you want to submit your same fix PR that you are doing here to that repo we would be happy to look at both.

GerbenWelter:
@RandomNoun7 Yeah I know, but I haven't used dsc_lite in that capacity. But I'll provide PRs for both.

GerbenWelter:
@RandomNoun7, as a first step I have created #410.

PR #410, opened 2019-04-24
(MODULES-8856) Fix using the dsc_psdscrunascredential parameter

GerbenWelter:
This PR fixes the regression introduced in #405 if the `dsc_psdscrunascredential` parameter is used by dsc_* resources in Puppet. If this was the case, the following was part of the Powershell code Puppet generated for DSC:

`password = [PSCustomObject]@{'user' = 'some_user'; 'password' = 'secret'} # PuppetSensitive | new-pscredential`

The use of '#' as the start of the comment breaks the parsing of the Powershell code because everything after the '#' was disregarded. Also the debug log wasn't redacted because the line didn't match.

This PR fixes this by using the multiline comment syntax on a single line which generates the code as following:

`password = [PSCustomObject]@{'user' = 'some_user'; 'password' = 'secret'} <# PuppetSensitive #> | new-pscredential`

Also the match was altered in two ways to account for this situation:
- Match the new  multiline comment suffix.
- Backreference anything before the sensitive value so we don't lose preceding key/value pairs in the redacted debug log because of the non-greedy match of preceding text.

This was needed otherwise the redacted version looked this:

`password = [PSCustomObject]@{'user' = '[REDACTED]' | new-pscredential`

### DIFF
--- a/lib/puppet/provider/base_dsc/powershell.rb
+++ b/lib/puppet/provider/base_dsc/powershell.rb
@@ -180,7 +180,7 @@ EOT
     when dsc_value.class.name == 'Hash'
       "@{" + dsc_value.collect{|k, v| format_dsc_value(k) + ' = ' + format_dsc_value(v)}.join('; ') + "}"
     when dsc_value.class.name == 'Puppet::Pops::Types::PSensitiveType::Sensitive'
-      "'#{escape_quotes(dsc_value.unwrap)}' # PuppetSensitive"
+      "'#{escape_quotes(dsc_value.unwrap)}' <# PuppetSensitive #>"
     else
       fail "unsupported type #{dsc_value.class} of value '#{dsc_value}'"
     end
@@ -192,12 +192,13 @@ EOT
 
   def self.redact_content(content)
     # Note that here we match after an equals to ensure we redact the value being passed, but not the key.
-    # This means a redaction of a string not including '= ' before the string value will not redact.
-    # Every secret unwrapped in this module will unwrap as "'secret' # PuppetSensitive" and, currently,
+    # This means a redaction of a string not including '= ' before the string value will not redact. We
+    # also match everything before the match and backreference it in the replacement string so we don't
+    # lose other key/value pairs on the same line.
+    # Every secret unwrapped in this module will unwrap as "'secret' <# PuppetSensitive #>" and, currently,
     # always inside a hash table to be passed along. This means we can (currently) expect the value to
     # always come after an equals sign.
-    # Note that the line may include a semi-colon and/or a newline character after the sensitive unwrap.
-    content.gsub(/= '.+' # PuppetSensitive;?(\\n)?$/,"= '[REDACTED]'")
+    content.gsub(/= ((?!' ; ').)+? <# PuppetSensitive #>/,"= '[REDACTED]'")
   end
 
   def ps_script_content(mode)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/) a
 
 ## [Unreleased]
 
+### Fixed
+
+- Ensure sensitive values are redacted in debug output and `psdscrunascredential` does not break ([MODULES-8856](https://tickets.puppetlabs.com/browse/MODULES-8856)) - thank you, [@Gerben Welter](https://github.com/GerbenWelter)!
+
 ## [1.9.1] - 2019-04-23
 
 ### Fixed
--- a/spec/unit/puppet/provider/powershell_spec.rb
+++ b/spec/unit/puppet/provider/powershell_spec.rb
@@ -39,17 +39,18 @@ describe Puppet::Type.type(:dsc_file).provider(:powershell) do
   describe "when secrets are present" do
     it "should unwrap secrets for passing to PowerShell" do
       sensitive_pass = Puppet::Pops::Types::PSensitiveType::Sensitive.new('password')
-      expect(subject.class.format_dsc_value(sensitive_pass)).to match(/'password' # PuppetSensitive/)
+      expect(subject.class.format_dsc_value(sensitive_pass)).to match(/'password' <# PuppetSensitive #>/)
     end
     it "should redact secrets for displaying in debug" do
       # Note that here we're passing a full string as it shows up in the script_content to be executed
       # This is because we built a matcher to redact the value being passed, but not the key.
       # This means a redaction of a string not including '= ' before the string value will not redact.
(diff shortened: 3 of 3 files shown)


michaeltlombardi:
Hey! This looks pretty good, but we'll also need to update the spec tests for it:

https://github.com/puppetlabs/puppetlabs-dsc/blob/master/spec/unit/puppet/provider/powershell_spec.rb#L39-L53

Investigating why CI failed (it failed before getting to the tests, but would eventually fail there too).

Thanks so much for pushing this up!

michaeltlombardi:
See failing tests here:

https://travis-ci.org/puppetlabs/puppetlabs-dsc/jobs/524116533#L2342-L2364

GerbenWelter:
Yeah, I was wondering why the tests failed there. I forgot to check for the unit tests. I will push a fix for that shortly.

michaeltlombardi:
Hey, @GerbenWelter! I took a peek locally and I noticed one small issue - when redacting a string it will only redact the _last_ secret on a line. For example (tested in interactive ruby 2.5.1):

```
irb(main):023:0> a
=> "@{'user' = 'someone' <# PuppetSensitive #> ; 'password' = 'S3cr3t' <# PuppetSensitive #>} | new-pscredential"
irb(main):024:0> b
=> "@{'user' = 'someone' ; 'password' = 'S3cr3t' <# PuppetSensitive #>} | new-pscredential"
irb(main):025:0> a.gsub(/(.*)= '.+' <# PuppetSensitive #>/,"\\1= '[REDACTED]'")
=> "@{'user' = 'someone' <# PuppetSensitive #> ; 'password' = '[REDACTED]'} | new-pscredential"
irb(main):026:0> b.gsub(/(.*)= '.+' <# PuppetSensitive #>/,"\\1= '[REDACTED]'")
=> "@{'user' = 'someone' ; 'password' = '[REDACTED]'} | new-pscredential"
```

This is a bit of an edge case though, this only _seems_ to happen if both the username and password are specified as sensitive in a declaration.

Thoughts, @ThoughtCrhyme, @RandomNoun7?

I'm half-inclined to merge this and circle back for the edge case, given this will get the module working again.

RandomNoun7:
@michaeltlombardi @GerbenWelter Agree'd. Merge now, get it out the door. Circle on the edge case later.

lib/puppet/provider/base_dsc/powershell.rb
```diff
@@ -192,12 +192,13 @@ def self.escape_quotes(text)
...
+    # This means a redaction of a string not including '= ' before the string value will not redact. We
+    # also match everything before the match and backreference it in the replacement string so we don't
+    # lose other key/value pairs on the same line.
     # Every secret unwrapped in this module will unwrap as "'secret' # PuppetSensitive" and, currently,
     # always inside a hash table to be passed along. This means we can (currently) expect the value to
-    # always come after an equals sign.
-    # Note that the line may include a semi-colon and/or a newline character after the sensitive unwrap.
-    content.gsub(/= '.+' # PuppetSensitive;?(\\n)?$/,"= '[REDACTED]'")
+    # always come after an equals sign and only exist once per line.
+    content.gsub(/(.*)= '.+' <# PuppetSensitive #>/,"\\1= '[REDACTED]'")
```
michaeltlombardi: Slightly more functional check via @Iristyle that ensures we redact any number of sensitive objects on a line:

```ruby
content.gsub(/= '.+?' <# PuppetSensitive #>/,"= '[REDACTED]'")
```

Iristyle:
I don't think that all of the capturing is needed in this case and we can take a simpler approach. If we know that code emitted for each sensitive value looks like `= 'secret' <# PuppetSensitive #>`, we can get away with a non-greedy match on the secret value like this:

```ruby
gsub(/= '.+?' <# PuppetSensitive #>/,"= '[REDACTED]'")`
```

But to handle secrets that themselves contain (escaped?) single quotes I think we have to make it even less strict and ignore the closing `'` on the non-greedy pattern like this:

```ruby
gsub(/= '.+? <# PuppetSensitive #>/,"= '[REDACTED]'")`
```

michaeltlombardi:
@Iristyle the second one seems to have swallowed the (meant to be unredacted) username and the key for password?

```
[18] pry(main)> b
=> "@{'user' = 'someone' ; 'password' = 'S3cr3t' <# PuppetSensitive #>} | new-pscredential"
[20] pry(main)> b.gsub(/= '.+? <# PuppetSensitive #>/,"= '[REDACTED]'")
=> "@{'user' = '[REDACTED]'} | new-pscredential"
```

Iristyle:
Ok, I was a little off on the first attempt, as I missed the `;` separated values.  I believe this should work fine by making sure to *not* capture a string containing `' ; '` in the non-greedy match. In other words, it will stop processing if it hits the boundary between parameters in the hash.


```
[31] pry(main)> a
=> "@{'user' = 'someone' <# PuppetSensitive #> ; 'password' = 'S3cr3t' <# PuppetSensitive #>} | new-pscredential"
[32] pry(main)> a.gsub(/= ((?!' ; ').)+? <# PuppetSensitive #>/,"= '[REDACTED]'")
=> "@{'user' = '[REDACTED]' ; 'password' = '[REDACTED]'} | new-pscredential"

[33] pry(main)> b
=> "@{'user' = 'someone' ; 'password' = 'S3cr3t' <# PuppetSensitive #>} | new-pscredential"
[34] pry(main)> b.gsub(/= ((?!' ; ').)+? <# PuppetSensitive #>/,"= '[REDACTED]'")
=> "@{'user' = 'someone' ; 'password' = '[REDACTED]'} | new-pscredential"
```

michaeltlombardi:
I tested this one, seems to function as expected!

As far as I can tell we should never get multiple secrets per line where the secrets are not separated by a semicolon.

lib/puppet/provider/base_dsc/powershell.rb
umaritimus: 👍

Iristyle:
Note that my regex `/= ((?!' ; ').)+? <# PuppetSensitive #>/` fails should the sensitive value contain `' ; '` but that seems pretty unlikely given escaping, etc.

michaeltlombardi:
Regex is _hard_.

michaeltlombardi:
@RandomNoun7, @ThoughtCrhyme, I've updated the PR to use the modified code per @Iristyle and added an additional test. I took a second to reword the commit and add a changelog entry crediting @GerbenWelter as well.

If CI passes we should probably get this merged and cut an immediate release.

GerbenWelter:
Thank you all! I have provided a similar fix for dsc_lite in puppetlabs/puppetlabs-dsc_lite#107.

PR #414, opened 2019-05-31
(FM-8072) convert acceptance tests to Litmus

ThoughtCrhyme:
In this commit we have added Litmus style acceptance tests that run in parallel with
the existing Beaker tests.

To call the tests run "bundle exec rspec ./spec/litmus" with an environment varialbe
of TARGET_HOST set to the target you wish to exercise the tests against.

Note: The parallel execution of tests via "bundle exec rake litmus:acceptance:parallel"
does not seem to be working at present.


spec/spec_helper_acceptance_litmus.rb
```diff
@@ -0,0 +1,127 @@
...
+  else
+    "'#{value.to_s.gsub("'","''")}\'"
+  end
+end
+
+def is_number?(value)
+  value.respond_to?(:to_i) ? value.to_s == value.to_i.to_s : false
+end
+
+def create_remote_windows_file(folder_path, file_name, content)
```
ThoughtCrhyme: This might be worth moving to Litmus as a helper method to upload files so we don't have to keep repeating ourselves, I stole the code to call bolt form Litmus here: https://github.com/puppetlabs/puppet_litmus/blob/master/lib/puppet_litmus/rake_tasks.rb#L180

spec/litmus/basic_dsc_resources/file/negative/file_invalid_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+
+      dsc_manifest = single_dsc_resource_manifest(dsc_type, dsc_props)
+
+      it_should_behave_like 'an error in a puppet run', dsc_manifest, /invalid value: cows/
+  end
+
+  context 'DSC File resource with invalid data for DestinationPath' do
+      dsc_props = {
+          :dsc_ensure          => 'Present',
+          :dsc_destinationpath => 'K:\\test.file',
```
ThoughtCrhyme: I haven't been very consistent about double slashes vs single slashes throughout.

spec/litmus/basic_dsc_resources/file/dir_valid_spec.rb
```diff
@@ -0,0 +1,122 @@
...
+
+describe 'Apply DSC "File" (directory) resource' do
+
+  dsc_type = 'file'
+  dsc_module = 'PSDesiredStateConfiguration'
+
+  context 'A Directory Resource with Valid "DestinationPath" Specified' do
+    before(:all) do
+      @dsc_destinationpath = 'C:\test_dir'
+      run_shell("remove-item -path C:/temp/test_dir -Force -Recurse -ErrorAction SilentlyContinue", :expect_failures => true)
```
ThoughtCrhyme: Litmus' "run_shell" method doesn't (yet?) have an option for "accept_all_exit_codes", so I've used "expect_failures" here, which is not semantically correct, I don't really expect failure, but I also don't care if there is one.

One solution is to add "accept_all_exit_codes" to Litmus' "run_shell" the same way I added "noop" to "apply_manifest" like in this commit:
https://github.com/puppetlabs/puppet_litmus/commit/5bcbfecb11761be094ce9be2679609db5754d9a2

spec/spec_helper_acceptance_litmus.rb
tphoney: these helper functions would be better suited moved out into a helper file.  ala https://github.com/puppetlabs/puppet_litmus/wiki/converting-tests-from-beaker-rspec-to-litmus#where-to-put-your-helper-functions have a look at https://github.com/puppetlabs/puppetlabs-postgresql/blob/master/spec/spec_helper_acceptance_local.rb

tphoney:
Adding in appveyor, to see if it works would be neato, ala https://github.com/puppetlabs/puppetlabs-motd/blob/master/appveyor.yml

spec/spec_helper_acceptance_litmus.rb
pmcmaw: Are there plans to move these helper functions to spec_helper_acceptance_local.rb @ThoughtCrhyme?

pmcmaw:
In general Im slightly put off the test duplication, however if there are plans to remove in the future and this is not forgotten about or it is at least tracked as a reminder this would be less worrying. Travis continues to run the tests in the usual way at what point or when will tests using litmus actually be running?

spec/spec_helper_acceptance_litmus.rb
ThoughtCrhyme: Will do that in subsequent commits when we remove the beaker tests.

ThoughtCrhyme:
@tphoney I could add to appveyor however, I'm not sure how many concurrent jobs we're allowed with our license, and I'm not sure we want bottleneck on appveyor seeing as how MOTD's acceptance tests are about 4x longer to run than the acceptance tests: https://ci.appveyor.com/project/puppetlabs/puppetlabs-motd

ThoughtCrhyme:
@pmcmaw yeah, I'm not stoked about test duplication either, but without a finalized plan for doing master/agent testing somewhere we still need beaker/jenkins (although we could probably remove the agent only tests, if those are being run in appveyor; although see the note above to TP about appveyor constraints).

glennsarti:
@ThoughtCrhyme It's currently 8 for the entire puppetlabs org in AppVeyor. Although it's up for renewal soon.

pmcmaw:
@ThoughtCrhyme I think that the acceptance tests taking 4x to run should outweight the fact that currently there are zero acceptance tests running on a PR basis. It will safe guard CI and the failures that will get through to jenkins going forward should be minimal.

appveyor.yml
```diff
@@ -48,9 +50,31 @@ environment:
...
+-
+  matrix:
+    only:
+      - ACCEPTANCE: yes
+  install:
+    - set PATH=C:\Ruby%RUBY_VERSION%\bin;%PATH%
+    - bundle install --jobs 4 --retry 2 --without build
+    - type Gemfile.lock
+  test_script:
+    - ps: Set-Content -Value 'modulepath = C:/projects/puppetlabs-dsc/spec/fixtures/modules' -Path 'C:\ProgramData\PuppetLabs\code\environments\production\environment.conf'
```
ThoughtCrhyme: Hey @tphoney I was able to get the module path in correctly by adding this (and line 11 above), but the powershell call that results from this line https://github.com/ThoughtCrhyme/puppetlabs-dsc/blob/FM-8072/spec/lib/dsc_utils_litmus.rb#L11 ends up with a newline character in it for some reason. See here: https://ci.appveyor.com/project/puppetlabs/puppetlabs-dsc/builds/25433595/job/m2oyl1ojldn5sgpl#L1230

I've tried stripping it out before forming the call, but it didn't work: https://github.com/puppetlabs/puppetlabs-dsc/pull/414/files#diff-58d69cbefccdc435228cb60aaf6b8552R6
and
https://github.com/puppetlabs/puppetlabs-dsc/pull/414/files#diff-58d69cbefccdc435228cb60aaf6b8552R10

ThoughtCrhyme: Ok, the stripping out of the newline now works, but now I'm getting "Test-Path was not expected at this time" https://ci.appveyor.com/project/puppetlabs/puppetlabs-dsc/builds/25459972/job/1id22a25pbnf30a7#L1231

First thought is this is a problem in run_shell, haven't debugged through it though, I'll check on Monday.

spec/spec_helper_acceptance_litmus.rb
```diff
@@ -0,0 +1,109 @@
...
+end
+
+def puppet_version
+  run_shell('puppet --version', :acceptable_exit_codes => 0).stdout
+end
+
+if ENV['TARGET_HOST'].nil? || ENV['TARGET_HOST'] == 'localhost'
+  puts 'Running tests against this machine !'
+  if Gem.win_platform?
+    set :backend, :cmd
```
ThoughtCrhyme: @tphoney : @RandomNoun7 and I found the problem. On windows localhost it's executing in cmd and not powershell, hence the failure of all the powershell calls :/

spec/lib/dsc_utils_litmus.rb
```diff
@@ -0,0 +1,125 @@
+def locate_dsc_module(expect_ps_failure)
+  # Init
+  module_paths = run_shell('puppet config print modulepath').stdout.gsub("\n", '').gsub("\r", '').split(';')
+  # Search the available module paths.
+  module_paths.each do |module_path|
+    dsc_module_path = "#{module_path}/dsc".gsub('\\', '/')
+    ps_command = "Test-Path -Type Container -Path '#{dsc_module_path}'"
+    run_shell("powershell.exe -NoProfile -Nologo -Command \"if ( #{ps_command} ) { exit 0 } else { exit 1 }\"", :expect_failures => expect_ps_failure) do |result|
```
ThoughtCrhyme: Feels like this idiom of `"powershell.exe -NoProfile -Nologo -Command \"<whatever Powershell you want to run>\""` (which we have to insert so we can properly call out to Powershell from the cmd prompt) should be moved to Litmus' `run_shell` so we don't have to continually repeat ourselves.

spec/spec_helper_acceptance_litmus.rb
RandomNoun7: In further research I'v found that it's true commands run locally are run in cmd, but not because of this setting. This setting belongs to `specinfra`, but we don't end up using it's `run_command` methods. If we used `specinfra`'s we would likely get powershell execution for free because even though the backend is called `cmd`, it actually [builds a powershell script](https://github.com/mizzy/specinfra/blob/master/lib/specinfra/backend/cmd.rb#L13) out of whatever you give to it.

Instead, specinfra is loaded first, but then bolt_spec/run is loaded later which has its own implementation of `run_command` which I think is clobbering `specinfra`'s.

In Bolt land you get a transport choice of `local`, `winrm` or `ssh`. When running commands against a remote node, you get a powershell session because Bolt [made that choice specifically when creating the session](https://github.com/puppetlabs/bolt/blob/master/lib/bolt/transport/winrm/connection.rb#L54). When executing commands on Windows locally they did not make the same choice, instead opting for a [simple call to Open3.capture3](https://github.com/puppetlabs/bolt/blob/master/lib/bolt/transport/local_windows.rb#L71).

We could potentially solve this in Litmus, but I believe the correct place to solve this may be in Bolt. It is actually Bolt that is introducing this inconsistency in behavior. Bolt should be providing a PowerShell wrapper for local command calls on Windows as PowerShell is the default environment most users will expect when running commandline commands on Windows.

RandomNoun7: I just created [BOLT-1429](https://tickets.puppetlabs.com/browse/BOLT-1429) in hopes the Bolt team will agree that this issue is best solved closer to the source, rather than in every individual consumer of `BoltSpec::Run.run_command`.

spec/spec_helper_acceptance_litmus.rb
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    file { "#{folder_path}/#{file_name}":
+      content => "#{content}",
+    }
+  FileManifest
+  apply_manifest(manifest)
+end
+
+def puppet_version
```
tphoney: this does not seem to be used anywhere.

spec/spec_helper_acceptance_litmus.rb
tphoney: nice catch, it might be worth throwing up a PR to show the bolt team what you want. It looks like they have similar behaviour to what we want for run_task and run_script, were it invokes the powershell helper methods.

ThoughtCrhyme:
Passing adhoc starting here: https://jenkins-master-prod-1.delivery.puppetlabs.net/view/modules/view/windows/view/_adhoc/view/dsc/job/forge-windows_puppetlabs-dsc_init-manual-parameters_adhoc/lastSuccessfulBuild/

.sync.yml
```diff
@@ -24,6 +24,13 @@ Gemfile:
       - gem: master_manipulator
       - gem: puppet-blacksmith
         version: '~> 3.4'
+      - gem: 'puppet_litmus'
```
tphoney: has pdksync been run recently ? 
the gem does not need to be added anymore, as it is in puppet-module-gems

.sync.yml
```diff
@@ -46,6 +53,8 @@ Gemfile:
 
 # The rakefile is heavily modified for the build process
 Rakefile:
+  requires:
+  use_litmus_tasks: true
```
tphoney: if pdksync has been run we do not need this line. if it hasnt it is needed.

spec/spec_helper_acceptance_litmus.rb
tphoney: are we dependant on this change going in ?

ThoughtCrhyme: We are not, it will run as long as we wrap all the calls with "powershell.exe -NoProfile -Nologo -Command ..." which is annoying, but functional. And we have a ticket here FM-8298 to remove that annoyance when Bill's BOLT-1429 gets merged.

spec/spec_helper_acceptance_litmus.rb
ThoughtCrhyme: Removed

.sync.yml
ThoughtCrhyme: How many places do we want to run the acceptance tests? They are already running in Appveyor, and again in Beaker. Do we want Puppet Pipelines runs in addition to all that?

ThoughtCrhyme: Removed reference to litmus gem

.sync.yml
ThoughtCrhyme: removed litmus reference

ThoughtCrhyme:
Passing adhoc starting here (well, it passed the smoke test and got stuck waiting for VM's in full acceptance, not sure if it passed that at time of writing): https://jenkins-master-prod-1.delivery.puppetlabs.net/view/modules/view/windows/view/_adhoc/view/dsc/job/forge-windows_puppetlabs-dsc_init-manual-parameters_adhoc/55/

lionce:
Hello @ThoughtCrhyme ,
For executing tests using litmus we use pipelines (https://pipelines.puppet.com/team-modules/builds/565871). I've created new application for this repository pointing to  ThoughtCrhyme:FM-8072 branch.

ThoughtCrhyme:
Added a distelli-manifest.yml and it passes pipelines against a single 2012r2wmf5 machine https://pipelines.puppet.com/team-modules/builds/566356

Can't call litmus:acceptance:parallel because our beaker tests are presently sitting in spec/acceptance.

Gemfile
```diff
@@ -35,6 +35,7 @@ group :system_tests do
   gem "beaker-testmode_switcher", '~> 0.4',           require: false
   gem "master_manipulator",                           require: false
   gem "puppet-blacksmith", '~> 3.4',                  require: false
+  gem "puppet_litmus",                                require: false, platforms: [:ruby, :mswin, :mingw, :x64_mingw] if ENV['PUPPET_GEM_VERSION'].nil? or ENV['PUPPET_GEM_VERSION'] !~ %r{ 5}
```
eimlav: This can be removed since litmus is now in puppet-module-gems.

metadata.json
```diff
@@ -39,7 +39,7 @@
     "desired-state-configuration",
     "dsc"
   ],
-  "pdk-version": "1.8.0",
-  "template-url": "https://github.com/puppetlabs/pdk-templates.git",
-  "template-ref": "heads/master-0-g7281db5"
+  "pdk-version": "1.10.0",
+  "template-url": "https://github.com/puppetlabs/pdk-templates.git#1.10.0",
+  "template-ref": "1.10.0-0-gbba9ac3"
```
eimlav: Due to changes introduced in pdk 1.10.0, using `pdk update` on a module using the old `template-url` syntax (a plain url) will default the target ref to whatever the latest release for the version of pdk you are using is. To fix this you need to run `pdk convert --template-url https://github.com/puppetlabs/pdk-templates --template-ref master`. This will add the `#master` ref to the url and using `pdk update` after this will work as expected. This is just a one off change basically.

provision.yaml
```diff
@@ -0,0 +1,6 @@
+default:
+  provisioner: vmpooler
+  images: ['win-2012r2-x86_64']
+release_checks:
+  provisioner: vmpooler
+  images: ['win-2012-x86_64', 'win-2012r2-x86_64', 'win-2016-x86_64', 'win-2019-x86_64']
```
eimlav: Although the parallel job isn't currently working, why do we not supply all supported OS for release_checks?

ThoughtCrhyme:
Ok, I've removed beaker and tests are passing in appveyor and in pipelines using litmus:acceptance:parallel https://pipelines.puppet.com/team-modules/builds/566635

provision.yaml
```diff
@@ -0,0 +1,6 @@
+default:
+  provisioner: vmpooler
+  images: ['win-2012r2-x86_64']
+release_checks:
+  provisioner: vmpooler
+  images: ['win-2012r2-wmf5-x86_64', 'win-2016-x86_64', 'win-2019-core-x86_64']
```
ThoughtCrhyme: @eimlav the reason we chose this set of machines to test against is for two reasons:
1) This is the set of machines that Jenkins tested against
2) Puppetlabs-dsc module requires Powershell 5

ThoughtCrhyme:
Passing pipelines in parallel here https://pipelines.puppet.com/team-modules/builds/566667

metadata.json
ThoughtCrhyme: pdk convert will nuke my changes in the appveyor.yaml file and try to do a rubocop check, which is part of a separate ticket: https://tickets.puppetlabs.com/browse/MODULES-9318

eimlav: Sounds good to me @ThoughtCrhyme .

provision.yaml
eimlav: Any particular reason we don't use the core images for this?

ThoughtCrhyme: nice catch, fixed it

PR #415, opened 2019-05-31
(MODULES-8171) Run fails if paths in $env:lib don't exist

carabasdaniel:
Prior to this commit, Puppet fails while compiling C# code within init_ps.ps1 with a cryptic no implicit conversion of nil into String message.

  This commit checks that the LIB environment variable contains invalid paths and fails with a useful error message.

### DIFF
--- a/lib/puppet_x/puppetlabs/powershell_manager.rb
+++ b/lib/puppet_x/puppetlabs/powershell_manager.rb
@@ -38,10 +38,24 @@ module PuppetX
         !win32console_enabled?
       end
 
+      def invalid_lib_paths?
+        if ENV["lib"].nil? || ENV["lib"].empty?
+          return false
+        else
+          ENV["lib"].split(";").each do |path|
+            if !File.directory?(path)
+              return true
+            end
+          end
+        end
+      end
+
       def initialize(cmd, debug, pipe_timeout)
         @usable = true
 
         named_pipe_name = "#{SecureRandom.uuid}PuppetPsHost"
+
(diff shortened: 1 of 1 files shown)


lib/puppet_x/puppetlabs/powershell_manager.rb
```diff
@@ -43,6 +43,10 @@ def initialize(cmd, debug, pipe_timeout)
 
         named_pipe_name = "#{SecureRandom.uuid}PuppetPsHost"
 
+        if !ENV["lib"].nil? && !File.exist?(ENV["lib"])
```
RandomNoun7: `lib` is a variable much like path which can be a delimited string of multiple directories. This check will need to split the string and iterate to check each path. If any of the paths are not valid the module will fail.

`directory?` should also probably be used here instead of `exist?`. If someone were to use a real file in this var, instead of a directory, then `exist?` would return true, but that would still be an invalid value for this variable.

lib/puppet_x/puppetlabs/powershell_manager.rb
```diff
@@ -38,11 +38,26 @@ def self.supported?
         !win32console_enabled?
       end
 
+      def check_valid_paths(paths)
```
RandomNoun7: If all we are doing is returning a boolean, we could conceivably change this to something like `def invalid_lib_paths?` and the call below could shorten to `raise <error message> if invalid_lib_paths?`

I also think that since `ENV['lib']`, as part of the environment, is necessarily global to all of the code in this file, there's no real need to pass it as a parameter. You can access it directly from within this function if the name is changed as above to make it clear that you are testing the `lib` variable specifically.

PR #418, opened 2019-06-25
(MODULES-7762) State Exact PS Version Requirement

RandomNoun7:
This is a documentation change to make it clear exactly what version
of PowerShell is required for the module to function. While 99% of
customers running WMF 5.0 will have a valid version, there was at least
one release that had a PowerShell version that will not support this
module and the DSC capabilities it requires.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/) a
 
 ## [Unreleased]
 
+### Added
+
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/) a
 
 ## [Unreleased]
 
+### Added
+
+- Added a note to the changelog specifying the exact PowerShell version required for the module to function ([MODULES-7762](https://tickets.puppetlabs.com/browse/MODULES-7762))
```
michaeltlombardi:
```suggestion
- Added a note to the readme specifying the exact PowerShell version required for the module to function ([MODULES-7762](https://tickets.puppetlabs.com/browse/MODULES-7762))
```

PR #430, opened 2019-10-22
(FM-8298) Cleanup Acceptance Spec commands

michaeltlombardi:
This commit removes the PowerShell prefixes to the run_shell calls
in the acceptance spec tests as Bolt now uses PowerShell as the
default shell for target nodes.


cmccrisken-puppet:
@michaeltlombardi If you want to rebase with upstream, we can probably get a successful CI run and merge this PR.

PR #434, opened 2019-11-20
MODULES-10117: Fix typo in xRDServer

BreadBomb:
Fix typo in the xRDServer powershell module that prevents the script from executing succesfully.


florindragos:
Hi @BreadBomb 
Thanks for the fix. Looks like this has also been fixed in the xRDServer module.

PR #465, opened 2020-11-17
pdksync - Update pdk-templates for gitpod and codespaces support

DavidS:
Update pdk-templates for gitpod and codespaces support
pdk version: `1.18.1`


DavidS:
ignoring this for now as this module is about to be deprecated.

